PR #24, opened 2025-02-17
Opt-in to specific secrets

coffee-cup:
This PR allows steps to opt-in to only invalidating when specific secret values change. For example, the node install step can invalidate only when variables that are prefixed with `NODE` change. As opposed to re-running when any secret changes. This is useful on platforms like Railway where a deployment id is passed in as a secret. With this PR, the install layer can be re-used even though that changes every build.

- Change the step `UseSecrets` to a `Secrets` array
- Specified a shared hint key on local state context
- Remove git and curl from the runtime image

### DIFF
--- a/core/generate/command_step_builder.go
+++ b/core/generate/command_step_builder.go
@@ -1,8 +1,10 @@
 package generate
 
 import (
+	"fmt"
 	"maps"
 
+	a "github.com/railwayapp/railpack/core/app"
 	"github.com/railwayapp/railpack/core/plan"
 	"github.com/railwayapp/railpack/core/utils"
 )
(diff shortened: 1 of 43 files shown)


core/generate/command_step_builder.go
```diff
@@ -15,7 +17,9 @@ type CommandStepBuilder struct {
 	Assets      map[string]string
 	Variables   map[string]string
 	Caches      []string
-	UseSecrets  bool
+	Secrets     *[]string
```
jaredLunde, line 20: Why not just `Secrets     []string`?

coffee-cup: Because I want to default to all secrets if this is nil when deserialized from a `railpack.json` file. If it is `[]string` then it will deserialize into an empty array

jaredLunde: Ah, yeah that's legit.

PR #56, opened 2025-03-05
Java provider

### LINKED ISSUE #51: Java LTS support
Hello, is Java considered in the most recent LTS versions, Spring Boot or other frameworks?

aleksrutins:
Closes #51 

Implementation mostly ported from Nixpacks' Java provider.

Currently includes just Java & Spring Boot support; my inclination is that other frameworks could probably be added in subsequent PRs.

Also adds a `ctx.NewMiseStepBuilder(displayName string)` public method (along with the existing private `newMiseStepBuilder`) to allow creation of new mise steps, e.g. for runtime packages.


core/providers/java/gradle.go
```diff
@@ -0,0 +1,105 @@
...
+const (
+	DEFAULT_GRADLE_VERSION = 8
+	GRADLE_CACHE_KEY       = "gradle"
+)
+
+func (p *JavaProvider) usesGradle(ctx *generate.GenerateContext) bool {
+	return ctx.App.HasMatch("gradlew")
+}
+
+func (p *JavaProvider) getGradleVersion(ctx *generate.GenerateContext) (int, error) {
```
coffee-cup, line 20: Instead of just returning a version, this should set the version for gradle on the context. That allows us to have "source" info which is displayed to the user. Basically lets them know why a specific version was used. [Deafult versions](https://railpack.com/architecture/package-resolution#previous-and-default-versions) are also treated bit differently than versions from files.

https://github.com/railwayapp/railpack/blob/main/core/providers/node/node.go#L241

screenshot-2025-03-05-13 57 56

core/providers/java/java.go
```diff
@@ -0,0 +1,72 @@
...
+
+		if ctx.App.HasMatch("mvnw") && !ctx.App.IsFileExecutable("mvnw") {
+			build.AddCommand(plan.NewExecCommand("chmod +x mvnw"))
+		}
+
+		build.AddCommand(plan.NewExecCommand(fmt.Sprintf("%s -DoutputFile=target/mvn-dependency-list.log -B -DskipTests clean dependency:list install", p.getMavenExe(ctx))))
+	}
+
+	ctx.Deploy.Inputs = []plan.Input{
+		plan.NewStepInput(build.Name()),
```
coffee-cup, line 55: With Railpack, there is a difference between the build and runtime images. The build is heavy and includes a lot of system dependencies for compiling packages (like build-essential). The runtime image is much smaller.

The deploy inputs should be based on the runtime image and just copy over things from the build (like the /app directory and mise executables).

See [this](https://github.com/railwayapp/railpack/blob/main/core/providers/python/python.go#L72) as an example.

core/generate/mise_step_builder.go
```diff
@@ -29,11 +29,11 @@ type MiseStepBuilder struct {
 	env                   *a.Environment
 }
 
-func (c *GenerateContext) newMiseStepBuilder() *MiseStepBuilder {
+func (c *GenerateContext) NewMiseStepBuilder(displayName string) *MiseStepBuilder {
```
coffee-cup, line 32: The way mise installs packages is by creating a file at `/etc/mise/config.toml`. Since the file is global, there can only be a single mise step builder (which is why it was private and didn't take a name previously)

core/providers/java/gradle.go
```diff
@@ -0,0 +1,92 @@
...
+	return ctx.App.HasMatch("gradlew")
+}
+
+func (p *JavaProvider) setGradleVersion(ctx *generate.GenerateContext) error {
+	miseStep := ctx.GetMiseStepBuilder()
+	gradle := miseStep.Default("gradle", DEFAULT_GRADLE_VERSION)
+
+	if envVersion, _ := ctx.Env.GetConfigVariable("GRADLE_VERSION"); envVersion != "" {
+		miseStep.Version(gradle, envVersion, "GRADLE_VERSION")
+	}
```
coffee-cup, line 25: The name is the second value returned in the tuple

```suggestion
	if envVersion, envName := ctx.Env.GetConfigVariable("GRADLE_VERSION"); envVersion != "" {
		miseStep.Version(gradle, envVersion, envName)
	}
```

core/providers/java/gradle.go
```diff
@@ -0,0 +1,92 @@
...
+
+	customVersion := string(versionRegex.FindSubmatch([]byte(wrapperProps))[3])
+
+	parseVersionRegex, err := regexp.Compile(`^(?:[\sa-zA-Z-"']*)(\d*)(?:\.*)(\d*)(?:\.*\d*)(?:["']?)$`)
+	if err != nil {
+		return err
+	}
+
+	if !parseVersionRegex.Match([]byte(customVersion)) {
+		return err
```
coffee-cup, line 53: I think this should just return `nil`
```suggestion
		return nil
```

core/providers/java/maven.go
```diff
@@ -0,0 +1,37 @@
...
+	if err != nil {
+		return ""
+	}
+
+	if strings.Contains(pomFile, "<groupId>org.wildfly.swarm") {
+		return "-Dswarm.http.port=$PORT"
+	} else if strings.Contains(pomFile, "<groupId>org.springframework.boot") &&
+		strings.Contains(pomFile, "<artifactId>spring-boot") {
+		return "-Dserver.port=$PORT"
+	}
```
coffee-cup, line 35: as someone who hasn't used the java ecosystem in a really long time, this is so confusing to me 🙃
Can you maybe add some comments (here and in the above places where lots of random files are read) that explains the decisions being made. Otherwise in a year it will be forgotton

core/providers/java/jdk.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/railwayapp/railpack/core/generate"
+)
+
+const DEFAULT_JDK_VERSION = "21"
+
+func (p *JavaProvider) setJDKVersion(ctx *generate.GenerateContext, miseOut *generate.MiseStepBuilder) error {
+	miseStep := ctx.GetMiseStepBuilder()
+	if miseOut == nil {
+		miseOut = miseStep
+	}
```
coffee-cup, line 15: I am a little confused about this. Why is `miseOut` needed? Can you just use the `miseStep` directly

core/providers/java/jdk.go
```diff
@@ -0,0 +1,28 @@
...
+	jdk := miseOut.Default("java", DEFAULT_JDK_VERSION)
+	if jdkVersion, _ := ctx.Env.GetConfigVariable("JDK_VERSION"); jdkVersion != "" {
+		miseOut.Version(jdk, jdkVersion, "JDK_VERSION")
+	}
```
coffee-cup, line 19:
```suggestion
	if jdkVersion, varName := ctx.Env.GetConfigVariable("JDK_VERSION"); jdkVersion != "" {
		miseOut.Version(jdk, jdkVersion, varName)
	}
```

core/providers/java/java.go
```diff
@@ -0,0 +1,87 @@
...
+
+	if p.usesGradle(ctx) {
+		p.setGradleVersion(ctx)
+		p.setJDKVersion(ctx, nil)
+
+		if ctx.App.HasMatch("gradlew") && !ctx.App.IsFileExecutable("gradlew") {
+			build.AddCommand(plan.NewExecCommand("chmod +x gradlew"))
+		}
+
+		build.AddCommand(plan.NewExecCommand("./gradlew clean build -x check -x test"))
```
coffee-cup, line 42: Is `test` needed when building an image? Builds will be faster if it is removed and railpack is not intended to be a CI platform.

core/providers/java/java.go
```diff
@@ -0,0 +1,87 @@
...
+		p.setJDKVersion(ctx, nil)
+
+		if ctx.App.HasMatch("mvnw") && !ctx.App.IsFileExecutable("mvnw") {
+			build.AddCommand(plan.NewExecCommand("chmod +x mvnw"))
+		}
+
+		build.AddCommand(plan.NewExecCommand(fmt.Sprintf("%s -DoutputFile=target/mvn-dependency-list.log -B -DskipTests clean dependency:list install", p.getMavenExe(ctx))))
+	}
+
+	runtimeDeps := ctx.NewMiseStepBuilder("mise:install-runtime")
```
coffee-cup, line 55: ah I see what the other mise builder is for. I guess this works? We could also use the `Exclude` part of the input to exclude the gradle binaries if they shouldn't be in the final image.

core/providers/java/java.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (p *JavaProvider) Plan(ctx *generate.GenerateContext) error {
+
+	build := ctx.NewCommandStep("build")
+	build.AddCommand(plan.NewCopyCommand("."))
+	build.Inputs = []plan.Input{plan.NewStepInput(ctx.GetMiseStepBuilder().Name())}
+
+	if p.usesGradle(ctx) {
+		p.setGradleVersion(ctx)
+		p.setJDKVersion(ctx, nil)
```
coffee-cup, line 36: Instead of passing in `nil`, can you always pass in the mise step that should be used so it is clear.

coffee-cup (commented):
Could you also please add a few tests for the java, gradle, jdk logic. Similar to the other providers. [Example](https://github.com/railwayapp/railpack/blob/main/core/providers/node/node_test.go)

core/generate/mise_step_builder.go
aleksrutins: I'm using the default one for build and a new one for runtime dependencies. Is there a better way to do this?

core/providers/java/jdk.go
aleksrutins: `miseOut` gets the runtime dependencies `MiseStepBuilder` for runtime - since I don't need Gradle or Maven for the runtime, just Java, it makes sense (to me) to have a separate mise step for that.

core/providers/java/java.go
aleksrutins: `-x check -x test` excludes `check` and `test`.

core/providers/java/jdk.go
coffee-cup: This makes sense. I mentioned it in another comment, but can you always pass in the miseStep (either for build or runtime deps). Having it optional and then defaulting to the build deps is a bit confusing imo

core/providers/java/java.go
```diff
@@ -0,0 +1,87 @@
...
+
+		if ctx.App.HasMatch("mvnw") && !ctx.App.IsFileExecutable("mvnw") {
+			build.AddCommand(plan.NewExecCommand("chmod +x mvnw"))
+		}
+
+		build.AddCommand(plan.NewExecCommand(fmt.Sprintf("%s -DoutputFile=target/mvn-dependency-list.log -B -DskipTests clean dependency:list install", p.getMavenExe(ctx))))
+	}
+
+	runtimeDeps := ctx.NewMiseStepBuilder("mise:install-runtime")
+	runtimeDeps.Inputs = []plan.Input{ctx.DefaultRuntimeInput()}
```
coffee-cup, line 56: I don't think this does anything. The inputs for the mise step should not really be changed since it needs the a specific input to get the mise binary

aleksrutins:
> Could you also please add a few tests for the java, gradle, jdk logic. Similar to the other providers. [Example](https://github.com/railwayapp/railpack/blob/main/core/providers/node/node_test.go)

Is there a difference between this and snapshots?

core/providers/java/gradle.go
```diff
@@ -0,0 +1,92 @@
...
+	if !versionRegex.Match([]byte(wrapperProps)) {
+		return err
+	}
+
+	customVersion := string(versionRegex.FindSubmatch([]byte(wrapperProps))[3])
+
+	parseVersionRegex, err := regexp.Compile(`^(?:[\sa-zA-Z-"']*)(\d*)(?:\.*)(\d*)(?:\.*\d*)(?:["']?)$`)
+	if err != nil {
+		return err
+	}
```
coffee-cup, line 50: I don't think these should error if regex fails. Instead it could add a warning (`ctx.Logger.Warn`) and return `nil`. Erroring will fail the build and we want to avoid that as much as possible. Much better to just use the default gradle version if we can't regex parse the properties file.

coffee-cup (commented):
Looking good! Just one small comment about not erroring if possible

PR #66, opened 2025-03-06
[trivial] Update docs to include the correct build command for building from source.

Excedrin:
Existing documentation suggests using ```go build -o railpack``` to build from source, which results in the error: ```no Go files in /path/to/railpack```. Adding `./cmd/...` seems to do the trick.


coffee-cup (approved):
thanks!

PR #95, opened 2025-03-11
Node: also use all files for preinstall scripts

aleksrutins:
I'm using a preinstall script to fetch and build a native dependency, for instance, and it needs my Makefile in order to do that.

### DIFF
--- a/core/providers/node/package_manager.go
+++ b/core/providers/node/package_manager.go
@@ -48,7 +48,7 @@ func (p PackageManager) installDependencies(ctx *generate.GenerateContext, packa
 	hasPostInstall := packageJson.Scripts != nil && packageJson.Scripts["postinstall"] != ""
 	hasPrepare := packageJson.Scripts != nil && packageJson.Scripts["prepare"] != ""
 
-	// If there is a postinstall script, we need the entire app to be copied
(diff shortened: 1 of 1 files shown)


core/providers/node/package_manager.go
```diff
@@ -44,12 +44,13 @@ func (p PackageManager) RunScriptCommand(cmd string) string {
 }
 
 func (p PackageManager) installDependencies(ctx *generate.GenerateContext, packageJson *PackageJson, install *generate.CommandStepBuilder) {
+	hasPreInstall := packageJson.Scripts != nil && packageJson.Scripts["preinstall"] != ""
 	hasPostInstall := packageJson.Scripts != nil && packageJson.Scripts["postinstall"] != ""
 	hasPrepare := packageJson.Scripts != nil && packageJson.Scripts["prepare"] != ""
 
 	// If there is a postinstall script, we need the entire app to be copied
 	// This is to handle things like patch-package
```
coffee-cup, line 52:
```suggestion
	// If there are any pre/post install scripts, we need the entire app to be copied
	// This is to handle things like patch-package
```

coffee-cup (approved):

PR #121, opened 2025-03-24
add single-command local dev setup

jaredLunde:
Ran into an issue where CONTRIBUTING.md was out-of-sync with docs. For ease-of-use, I think we should try to keep local development setup constrained to a single command.

- Adds a default environment variable for `BUILDKIT_HOST` to mise config
- Automatically create the buildkit host or silently fail if it's a duplicate, etc.
- Add shortcut for running the CLI from mise


coffee-cup (approved):
🙌

PR #122, opened 2025-03-24
add ruby provider

jaredLunde:
Add Ruby provider with similar test coverage to Nixpacks. Includes:

- Rails
- Node provider
- ExecJS
- YJIT
- Bootsnap
- Sprockets
- Postgres/MySQL/Magick/Vips/etc


core/providers/ruby/ruby.go
```diff
@@ -0,0 +1,392 @@
...
+
+	if gemfileVersion := parseVersionFromGemfile(ctx); gemfileVersion != "" {
+		miseStep.Version(ruby, gemfileVersion, "Gemfile")
+	}
+
+	miseStep.AddSupportingAptPackage("libyaml-dev")
+	version := p.getRubyVersion(ctx)
+	version = utils.ExtractSemverVersion(version)
+	semver, err := utils.ParseSemver(version)
+	// YJIT in Ruby 3.1+ requires rustc to install
```
coffee-cup, line 255: wat

internal/utils/utils.go
```diff
@@ -108,3 +110,47 @@ func ExtractSemverVersion(version string) string {
 	}
 	return ""
 }
+
+// ParseSemver parses a semantic version string and returns a Semver struct.
+func ParseSemver(version string) (*Semver, error) {
```
coffee-cup, line 115: nit: add some tests for this

coffee-cup (approved):
Looks good!

core/providers/ruby/ruby.go
nbw: 🙃 YJIT is a rust based JIT that Shopify has been driving. It replaces the internal one written in C. It's good that we support it because it's becoming more adopted.

PR #123, opened 2025-03-26
add rust provider

jaredLunde:
- use correct build architecture
- select bins
- works with workspaces
- rocket works


cli/build.go
```diff
@@ -49,6 +49,12 @@ var BuildCommand = &cli.Command{
 		},
 	}, commonPlanFlags()...),
 	Action: func(ctx context.Context, cmd *cli.Command) error {
+		platform, err := getPlatform(cmd.String("platform"))
+		if err != nil {
+			return cli.Exit(err, 1)
+		}
+		os.Setenv("TARGETPLATFORM", platform.String())
```
jaredLunde, line 56: this is a bad idea(tm) going to think of an alternative solution to passing the platform through. we have to be aware of this in the plan afaict so I'll have to figure out how to get it into the build context I guess.

core/providers/rust/rust.go
```diff
@@ -0,0 +1,693 @@
...
+
+	return nil
+}
+
+func (p *RustProvider) StartCommandHelp() string {
+	return "To start your Rust application, Railpack will look for:\n\n" +
+		"1. A main.ts, main.js, main.mjs, or main.mts file in your project root\n\n" +
+		"2. If no main file is found, it will use the first .ts, .js, .mjs, or .mts file found in your project\n\n" +
+		"The selected file will be run with `rust run --allow-all`"
+}
```
coffee-cup, line 82: This seems to be incorrect

core/providers/rust/rust.go
```diff
@@ -0,0 +1,693 @@
...
+		workspace := p.resolveCargoWorkspace(ctx)
+
+		if workspace != "" {
+			return fmt.Sprintf("./bin/%s", workspace)
+		}
+
+		return p.getStartBin(ctx)
+	}
+
+	workspace := p.resolveCargoWorkspace(ctx)
```
coffee-cup, line 97: This can only be read once and moved to the top of the file

core/providers/rust/rust.go
```diff
@@ -0,0 +1,693 @@
...
+
+	if len(bins) == 1 {
+		bin = bins[0]
+	} else if envBinName, _ := ctx.Env.GetConfigVariable("RUST_BIN"); envBinName != "" {
+		for _, b := range bins {
+			if b == envBinName {
+				bin = b
+				break
+			}
+		}
```
coffee-cup, line 126: We probably should warn with `ctx.Logger.Warn` if `RAILPACK_RUST_BIN` is provided but not found

core/providers/rust/rust.go
```diff
@@ -0,0 +1,693 @@
...
+	if p.shouldUseMusl(ctx) {
+		return getRustTarget(ctx)
+	}
+
+	return ""
+}
+
+func (p *RustProvider) Install(ctx *generate.GenerateContext, install *generate.CommandStepBuilder) {
+	ctx.Caches.AddCache("cargo_registry", CARGO_REGISTRY_CACHE)
+	ctx.Caches.AddCache("cargo_git", CARGO_GIT_CACHE)
```
coffee-cup, line 157: The caches need to be added to the specific step. Something like

```go
install.AddCache(ctx.Caches.AddCache("cargo_registry", CARGO_REGISTRY_CACHE))
```

_Realizing the ergonomics here are a bit strange. Lmk if you have any suggestions_

core/providers/rust/rust.go
```diff
@@ -0,0 +1,693 @@
...
+	}
+
+	// Don't do anything if we're in a workspace for now
+	if workspace := p.resolveCargoWorkspace(ctx); workspace != "" {
+		return
+	}
+
+	install.AddCommands([]plan.Command{
+		plan.NewExecCommand(`mkdir -p src`),
+		plan.NewExecShellCommand(dummyCmd),
```
coffee-cup, line 184: We can pass in a custom name here so it shows up as something more useful then the `echo...` in the logs. We could also use a `NewFileCommand` with an asset to write the file too. Although the end result would be the same.

https://github.com/railwayapp/railpack/blob/main/core/core.go#L174

core/providers/rust/rust.go
```diff
@@ -0,0 +1,693 @@
...
+		// Check for "openssl" in any of the dependency maps
+		if _, ok := cargoToml.Dependencies["openssl"]; ok {
+			return true
+		}
+		if _, ok := cargoToml.DevDependencies["openssl"]; ok {
+			return true
+		}
+		if _, ok := cargoToml.BuildDependencies["openssl"]; ok {
+			return true
+		}
```
coffee-cup, line 405: Extract to `usesDep` function? I execpt it will be used for other checks too

cli/common.go
```diff
@@ -61,6 +61,14 @@ func GenerateBuildResultForCommand(cmd *cli.Command) (*core.BuildResult, *a.App,
...
 		return nil, nil, nil, fmt.Errorf("error creating env: %w", err)
 	}
+	// Make the build platform available to the build plan
+	if env.Variables["TARGETPLATFORM"] == "" {
+		platform, err := getPlatform(cmd.String("platform"))
+		if err != nil {
+			return nil, nil, nil, cli.Exit(err, 1)
+		}
+		env.Variables["TARGETPLATFORM"] = platform.String()
+	}
```
coffee-cup, line 71: Thoughts on passing it in the options and then having on the generate context? This works too, but I think it might be a bit confusing having `os.Getenv` in the providers

coffee-cup (commented):
Having musl support is nice, but Nixpacks mainly did that because it would run in an alpine image if possible. For Railpack, at least for now, all the runtime images are debian-slim. Do we need musl support in this case? Binary likely is smaller, but it might simplify things to always build with glibc?

cli/common.go
jaredLunde: Yeah I can put in there. I was between those two options.

jaredLunde:
> Having musl support is nice, but Nixpacks mainly did that because it would run in an alpine image if possible. For Railpack, at least for now, all the runtime images are debian-slim. Do we need musl support in this case? Binary likely is smaller, but it might simplify things to always build with glibc?

@coffee-cup Yep we can get rid of it. I assumed it was for smaller binary reasons so I didn't want to undo that unilaterally.

core/providers/rust/rust.go
jaredLunde: I don't think it's necessary if we don't do musl : )

core/providers/rust/rust.go
jaredLunde: you sure?

cli/common.go
jaredLunde: Actually I think we can maybe remove this entirely if (again) we drop musl

PR #126, opened 2025-03-27
add dotnet provider

jaredLunde:
- cache `dotnet restore`
- include binary and dotnet root in runtime image (the binary needs the dotnet runtime)
- add internationalization apt package by default


iloveitaly:
@jaredLunde this looks like an awesome change and I'd love to get it merged in. Would you mind rebasing on master? Thanks!

jaredLunde:
blegh, struggling to update snapshots on the hotel wifi for some reason

danjrwalsh:
Thanks for your work on this, excited to see this get merged!

Quick question - any reason why 6 is the default version? I know it was the default version in Nixpacks but .NET 6 has been EOL since Nov 2024. .NET 8 has been the most recent LTS for almost 2 years now, which itself is soon to be replaced by .NET 10 in 2 months. Probably makes sense to have the most recent LTS be the default version?

jaredLunde:
> Thanks for your work on this, excited to see this get merged!
> 
> Quick question - any reason why 6 is the default version? I know it was the default version in Nixpacks but .NET 6 has been EOL since Nov 2024. .NET 8 has been the most recent LTS for almost 2 years now, which itself is soon to be replaced by .NET 10 in 2 months. Probably makes sense to have the most recent LTS be the default version?

@danjrwalsh It's been 6 months now but if I had to guess I copy/pasted from Nixpacks. Can update.

examples/dotnet-api/Controllers/WeatherForecastController.cs
```diff
@@ -0,0 +1,32 @@
+using Microsoft.AspNetCore.Mvc;
+
+namespace csharp.Controllers;
+
+[ApiController]
+[Route("[controller]")]
+public class WeatherForecastController : ControllerBase
```
iloveitaly: it doesn't look like this is used in the app entrypoint, can we remove?

core/providers/dotnet/dotnet.go
```diff
@@ -0,0 +1,205 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/railwayapp/railpack/core/generate"
+	"github.com/railwayapp/railpack/core/plan"
+	"github.com/railwayapp/railpack/internal/utils"
+)
+
+const (
+	DEFAULT_DOTNET_VERSION   = "6.0.428"
```
iloveitaly: Could we bump this to the latest LTS version?

core/providers/dotnet/dotnet.go
```diff
@@ -0,0 +1,205 @@
...
+		plan.NewCopyCommand("global.json*"),
+		plan.NewExecCommand(fmt.Sprintf("mkdir -p %s", DOTNET_DEPENDENCIES_ROOT)),
+		plan.NewExecCommand(`dotnet restore`),
+	})
+}
+
+func (p *DotnetProvider) Build(ctx *generate.GenerateContext, build *generate.CommandStepBuilder) {
+	maps.Copy(build.Variables, p.GetEnvVars(ctx))
+	build.AddCommands([]plan.Command{
+		plan.NewCopyCommand("."),
```
iloveitaly: [we should use NewLocalLayer](https://github.com/railwayapp/railpack/blob/afb0ac970de386a0b8de0a9a077d132d0ec79b7a/core/providers/python/python.go#L73) which ensures that `.dockerignore` is respected

iloveitaly (changes requested):
@coffee-cup left my comments on the PR. All of them are pretty minor—I think we should merge and I can handle all of the comments in a follow-up PR.

PR #127, opened 2025-03-28
add elixir provider

jaredLunde:
- autodetect OTP version
- multistage: install, compile deps in a separate step from build
- compiles to binary unlike nixpacks which runs `mix phx.server`. the mise installs do not wind up in the final image.

Results in much smaller image than nixpacks. Leverages docker caching way better as well.

Screenshot 2025-03-27 at 8 34 46 PM


core/providers/elixir/elixir.go
```diff
@@ -0,0 +1,233 @@
...
+	case "1.0", "1.1":
+		return "18"
+	case "1.2", "1.3":
+		return "19"
+	case "1.4":
+		return "20"
+	case "1.5":
+		return "20"
+	case "1.6":
+		return "2`"
```
FugiTech, line 215: This looks like a typo? I assume it should be `21`

FugiTech (commented):
This is awesome, I hope it can be added in soon!

I'm not sure how much specific phoenix support makes sense, but afaik they encourage you to install npm dependencies with `npm install --prefix assets <packages>`, which you can then import in your `/assets/app.js`. This means the build would have to download those npm packages or the build will break.

It would be nice if this could detect the presence of `assets/package.json` (and/or `assets/package-lock.json`) and if so include `node` as a package dependency and run `npm ci --prefix assets` in the install step.

core/providers/elixir/elixir.go
jaredLunde: Yep! Thank you for reviewing

jaredLunde:
@FugiTech the `mix assets.deploy` step should install NPM dependencies, no?

PR #133, opened 2025-03-31
List all languages on "Getting Started" docs

laggycomputer:
Momentarily got confused on whether Rust was supported since the Getting Started page seemed not to acknowledge the announcement in the recent newsletter that Rust was added.

Properly added every supported language to the list on the Getting Started docs page.

### DIFF
--- a/docs/src/content/docs/getting-started.mdx
+++ b/docs/src/content/docs/getting-started.mdx
@@ -18,6 +18,7 @@ the docs for each language that Railpack supports.
 - [Deno](languages/deno)
 - [Rust](languages/rust)
 - [Ruby](languages/ruby)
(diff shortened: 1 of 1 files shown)


laggycomputer:
> Adding Rust and Ruby is good, but why were these re-ordered? Node and Python are by far the most popular, which is why they were at the top.

I intended the order on the page to be consistent with the order in the sidebar. I'm not very opinionated and if you believe it should be some rough popularity order then I'll make that change.

coffee-cup:
Could you keep the order from before for now. Can slot in Rust and Ruby after PHP.

coffee-cup (approved):

coffee-cup:
Thanks!

PR #147, opened 2025-04-23
add 404 page in docs `Caddyfile`

Fredkiss3:
## Summary 
Use the dedicated 404 page instead of redirecting to index page on 404 errors.

## Why ?

As of now, when you go on an [nonexistent](https://railpack.com/nonexistent) page in the railpack docs website , you get redirected to the home page, while there is a dedicated 404 page for the website. 

## How ?
- I updated the Caddyfile to show the custom 404 page whenever there is a not found error. 
- I also removed the last `index.html` in the `try_files` instruction as it would always fall back to the index page in that case. this is actually good for SPAs, but the docs website is not a SPA.

### DIFF
--- a/docs/Caddyfile
+++ b/docs/Caddyfile
@@ -49,5 +49,12 @@
 
 	file_server
 
-	try_files {path} {path}.html {path}/index.html /index.html
+	try_files {path} {path}.html {path}/index.html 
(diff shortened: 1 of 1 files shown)


coffee-cup (approved):
thanks 🙌

PR #151, opened 2025-05-01
Reorder mise package version detection for Elixir

FugiTech:
When detecting which version of Elixir & Erlang to install, it _seemed_ like the order was reversed. I would expect an explicit environment variable to have the highest priority, then `.elixir-version`, and finally falling back to `mix.exs` or the default as a last resort. Instead, it would always use the value in `mix.exs` if available, since the last call to `.Version` wins.

I say "seemed" because I notice that most providers have this potentially incorrect ordering. At a glance, `gradle`, `node`, `python`, `ruby`, and `rust` all do not treat the environment variable as highest priority. If this PR is merged it may warrant addressing those providers as well, but I did not have confidence in my ability to choose the correct ordering for those ecosystems.

### DIFF
--- a/core/providers/elixir/elixir.go
+++ b/core/providers/elixir/elixir.go
@@ -115,22 +115,21 @@ var elixirVersionRegex = regexp.MustCompile(`(elixir:[\s].*[> ])([\w|\.]*)`)
 func (p *ElixirProvider) InstallMisePackages(ctx *generate.GenerateContext, miseStep *generate.MiseStepBuilder) {
 	elixir := miseStep.Default("elixir", DEFAULT_ELIXIR_VERSION)
 
-	if envVersion, varName := ctx.Env.GetConfigVariable("ELIXIR_VERSION"); envVersion != "" {
-		miseStep.Version(elixir, envVersion, varName)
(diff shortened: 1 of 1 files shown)


jaredLunde:
Hmm I've noticed this myself and I hear what you are saying but shouldn't this be desirable?
> it would always use the value in `mix.exs` if available

As in, this would all but ensure the version is correct for the application?

FugiTech:
Using the value in `mix.exs` is likely _good_ but I don't think it is the _best_. My expectation is that most people have it in the form of `elixir: "~> 1.15"` which is just setting a lower bound of 1.15, not wanting exactly 1.15. Whereas if `.elixir-version` exists I would expect that to be an explicit version number that is exactly what the developer wants, as the developer had to put in extra effort to create that file and specify a version.

Finally, as a developer I'd expect if I'm manually setting `RAILPACK_ELIXIR_VERSION` then it should always be used, since I've gone out of my way to set it. If it were something that be overwritten by `mix.exs` or `.elixir-version` then I would expect it to be `RAILPACK_DEFAULT_ELIXIR_VERSION` to indicate it is just a default value that might not be respected.

jaredLunde:
Nice, all very points. Will try this locally and then merge 😄

PR #153, opened 2025-05-02
Install node dependencies in phoenix projects

FugiTech:
Phoenix offers [three methods of installing JS dependencies](https://hexdocs.pm/phoenix/asset_management.html#third-party-js-packages). One of them is to add a `package.json` to the `assets/` folder. If a developer uses that method, the build system needs to run `npm install` or `mix assets.deploy` will fail.

This change checks if `assets/package.json` exists and if so adds `node` to the mise packages, and then injects a new command step to run `npm install` (or `npm ci` if `package-lock.json` exists) and modifies the build step to depend on it. I tried to ensure this is performant as possible by allowing it to run in parallel to installing elixir dependencies and utilizing caching the same way the node provider does.

I tried to make this change as simple as possible by not detecting a specific version of node, nor supporting alternate package managers like yarn, pnpm or bun. Since this only runs in the build step and only downloads a bunch of files into `node_modules` I expect the specific version of node doesn't matter too much, and I was unsure how common it is to use alternative package managers in Phoenix so did not know if it was worth the effort. However it shouldn't be too hard to add in the future if needed.

### DIFF
--- a/core/providers/elixir/elixir.go
+++ b/core/providers/elixir/elixir.go
@@ -7,8 +7,10 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/railwayapp/railpack/core/app"
 	"github.com/railwayapp/railpack/core/generate"
 	"github.com/railwayapp/railpack/core/plan"
+	"github.com/railwayapp/railpack/core/providers/node"
 	"github.com/railwayapp/railpack/internal/utils"
 )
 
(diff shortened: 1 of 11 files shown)


jaredLunde:
Hey thanks for all of the work you're putting into the Elixir provider! These changes are great. I think for the `npm install` step we'll want to do something similar to what we do with [Ruby](https://github.com/railwayapp/railpack/blob/main/core/providers/ruby/ruby.go#L44-L78) and [PHP](https://github.com/railwayapp/railpack/blob/main/core/providers/php/php.go#L179)

I'd be happy to help you out with this if needed!

FugiTech:
Thanks! I hadn't looked at the Ruby and PHP implementations, thank you for pointing them out. I see they re-use the Node provider, which does seem like a much better approach.

Unfortunately, for Phoenix the `package.json` is not in the root of the app but under `assets/`. It looks like the Node provider has a lot of things hardcoded to the app root, and I'm not sure the best way to handle making it work with `assets/package.json`. I think it would solve _some_ of the issues if we temporarily replace `ctx.App` with one set to the `assets` folder, but I think there would still be some problems. Alternatively, I could add a `Prefix` field to the Node provider and inject it into all paths, but I'm worried that might be too invasive.

I'll continue to think about it, and would appreciate it if you have any ideas!

FugiTech:
I found a way to make it work by replacing `ctx.App` and modifying the copy steps that the node provider makes. I'm not thrilled with the approach, but it minimizes changes to other parts of Railpack.

jaredLunde:
I think this approach is totally fine! Thanks for making that change.

core/providers/elixir/elixir.go
```diff
@@ -60,6 +62,44 @@ func (p *ElixirProvider) Plan(ctx *generate.GenerateContext) error {
 	})
 	ctx.Deploy.StartCmd = p.GetStartCommand(ctx)
 
+	// Node (if necessary)
+	if app, err := app.NewApp(ctx.App.Source + "/assets"); err == nil {
+		// All providers assume they're running in the application root
+		// but Phoenix puts it in the assets folder, so we have to lie to the provider
+		ctx.App = app
```
jaredLunde, line 69: For safety I wonder if we want an `originalApp` var that we can reset on the context e.g.

```go
originalApp := ctx.app
ctx.App = app
...
ctx.App = originalApp
```

jaredLunde: Otherwise looks good to me

FugiTech: With the current codebase it doesn't matter since `ctx.App` is copied into places that use it, but you're right that it might not always be that way and the change should be cleaned up. I'll make that change, and move the logic into a function so I can use `defer` and reduce some of the nested `if`s.

PR #157, opened 2025-05-15
No secrets for Elixir install step

FugiTech:
As far as I'm aware, there are no useful environment variables that can be set when installing dependencies for an Elixir app. Technically, [Mix does accept some](https://hexdocs.pm/mix/Mix.html#module-environment-variables) but they are to configure paths when building, which don't make sense for users to override when building inside Railpack.

By removing all secrets we ensure the cache won't get busted, significantly speeding up builds.

### DIFF
--- a/core/__snapshots__/TestGenerateBuildPlanForExamples_elixir-ecto_1.snap.json
+++ b/core/__snapshots__/TestGenerateBuildPlanForExamples_elixir-ecto_1.snap.json
@@ -93,9 +93,6 @@
     }
    ],
    "name": "install",
-   "secrets": [
-    "*"
-   ],
    "variables": {
     "ELIXIR_ERL_OPTIONS": "+fnu",
     "LANG": "en_US.UTF-8",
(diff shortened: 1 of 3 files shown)


FugiTech:
This is mostly focused on speeding up builds on Railway, where (I believe) the automatic GIT_SHA variables break the cache currently. Sadly I don't have enough insight into how Railway uses Railpack to further optimize. It seems like sometimes the mise step just gets uncached for no reason and erlang/elixir have to be installed from scratch which takes a while, which would then require deps to get installed and compiled too. Hopefully someone who works for Railway can improve that.

In my research it looks like it is common for CI systems to cache `deps/`, `_build/` and (sometimes) `~/.hex`. I was unsure if that's really needed and what the downsides of adding more cache folders would be. Again, I hope someone else can validate if those are needed and add them if so.

Milanhe92:
validate-github-github-docs-urls.yml.txt

PR #193, opened 2025-07-03
Fix ruby version detection

gauravtiwari:
This occurred because the regex pattern for extracting Ruby version constraints from Gemfiles was too restrictive and only matched single version numbers, not multi-part constraints like `ruby '>= 3.2.0', '< 3.5.0'`.

This fix enables Railpack to properly handle Ruby applications with version constraints in their Gemfiles, such as [Mastodon](https://github.com/mastodon/mastodon/blob/main/Gemfile) and other modern Ruby applications that specify version ranges rather than exact versions.

## Problem
The original regex pattern `ruby (?:'|")(.*)(?:'|")[^>]"` was designed to match single version strings but failed to properly extract complex version constraints that are common in modern Ruby applications.

## Fix
1. **Fixed regex pattern**: Updated `gemfileVersionRegex` to `ruby\s+(?:'|")([^'"]+)(?:'|")` to properly capture full version constraints
2. **Enhanced version filtering**: Added logic to skip preview/dev/rc versions when resolving constraints, ensuring only stable Ruby versions are selected
3. **Improved constraint resolution**: Integrated with Mise's `GetAllVersions` method and `github.com/Masterminds/semver/v3` to dynamically resolve the highest version satisfying the constraint
4. **Updated priority order**: Ensured constraint-based version resolution takes precedence over other version sources

## Tests
- Added comprehensive tests using real example projects

### DIFF
--- a/core/providers/ruby/ruby.go
+++ b/core/providers/ruby/ruby.go
@@ -5,8 +5,10 @@ import (
 	"fmt"
 	"maps"
 	"regexp"
+	"sort"
 	"strings"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/railwayapp/railpack/core/generate"
 	"github.com/railwayapp/railpack/core/plan"
 	"github.com/railwayapp/railpack/core/providers/node"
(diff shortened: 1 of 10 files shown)


core/providers/ruby/ruby.go
```diff
@@ -245,8 +247,28 @@ func (p *RubyProvider) InstallMisePackages(ctx *generate.GenerateContext, miseSt
 		miseStep.Version(ruby, utils.ExtractSemverVersion(string(versionFile)), ".ruby-version")
 	}
 
-	if gemfileVersion := parseVersionFromGemfile(ctx); gemfileVersion != "" {
-		miseStep.Version(ruby, gemfileVersion, "Gemfile")
+	if constraint := extractRubyConstraintFromGemfile(ctx); constraint != "" {
```
coffee-cup, line 250: Can you please extract this check into a separate function

core/providers/ruby/ruby.go
```diff
@@ -245,8 +247,28 @@ func (p *RubyProvider) InstallMisePackages(ctx *generate.GenerateContext, miseSt
...
+		allVersions, err := miseStep.Resolver.GetMise().GetAllVersions("ruby", "")
```
coffee-cup, line 251: This is not exactly how the resolver was intended to be used, but I think it is okay?

It might actually be cleaner to use [`b.Resolver.SetVersionAvailable`](https://github.com/railwayapp/railpack/blob/e28b250f38f3071dfcf431b2ca665a5ae6729b31/core/resolver/resolver.go#L88) instead of fetching and comparing the latest version here. It would be something like this

```go
miseStep.SetVersionAvailable(ruby, func(version string) bool {
    return !(strings.Contains(version, "-dev") || strings.Contains(version, "-preview")
})
```

This will better hook into the existing way Railpack checks for available versions

iloveitaly:
@gauravtiwari can you rebase this on master or give me edit privileges on the PR so I can resolve the snapshot conflicts?

@coffee-cup this PR LGTM once merge conflicts are handled—anything else you want done here?

coffee-cup:
This looks good once conflicts are resolved

gauravtiwari:
Hi Michael & Jake - This is ready for another pass and merge. Thanks!

PR #205, opened 2025-07-15
java: Add production build flag

aleksrutins:
This is used in Spring Boot to, for instance, disable the Vaadin debug server (which will break at runtime if it can't find a source directory, which is not provided at runtime).


aleksrutins:
CI failure seems to be a random GitHub ratelimit, unrelated to the PR.

coffee-cup (commented):
Change looks good. Can you please update the snapshots

aleksrutins:
GitHub ratelimit again.

PR #221, opened 2025-08-16
react router v7

iloveitaly:
As part of the dockerignore work, I realized that react router SPA was not being detected properly. This adds support for react router SPA *and* non-SPA mode.

At some point, we'll want to have the ability to build and start a container, then hit a URL on that container, to verify it built properly. There were a couple
of errors I ran into only on startup that won't be caught by the tests. I've added this to my list.


iloveitaly:
Pretty sure the test is failing here due to #216 or an intermittent outage. Going to wait until the mise upgrade is merged to retest this.

examples/node-vite-react-router-spa/generate.sh
```diff
@@ -0,0 +1,9 @@
+# Description: autogen this repo from the latest version of the RR template
```
iloveitaly, line 1:
```suggestion
#!/usr/bin/env zsh

cd "${0:A:h}"
# Description: autogen this repo from the latest version of the RR template
```

PR #222, opened 2025-08-16
add schema to plan

iloveitaly:
Makes it easier for users to debug what is wrong and generally understand the plan layout.

I ran into when working on another bug which resulted in an invalid railpack.json. I had to manually
add `$schema` to figure out what is going on.


cli/build.go
```diff
@@ -84,7 +89,6 @@ var BuildCommand = &cli.Command{
 
 		err = buildkit.BuildWithBuildkitClient(app.Source, buildResult.Plan, buildkit.BuildWithBuildkitClientOptions{
 			ImageName:    cmd.String("name"),
-			DumpLLB:      cmd.Bool("llb"),
 			OutputDir:    cmd.String("output"),
```
iloveitaly, line 92:
```suggestion
			DumpLLB:      cmd.Bool("llb"),
			OutputDir:    cmd.String("output"),
```

Issue #204, opened 2025-07-14
Caddyfile is not autogenerated for SPA applications


iloveitaly:
Here's how to reproduce:

```
git clone http://github.com/iloveitaly/python-starter-template
cd python-starter-template
just build_javascript
```

Here's the build output

```
╭────────────────╮
│ Railpack 0.2.1 │
╰────────────────╯

  ⚠ Failed to read config file `railpack.json`
  Use the following schema to validate your config file: https://schema.railpack.com

  ↳ Detected Node
  ↳ Using pnpm package manager
  ⚠ Node_modules directory found in project root, this is likely a mistake
  ⚠ It is recommended to add node_modules to the .gitignore file

  Packages
  ──────────
  pnpm  │  9.15.9   │  railpack default (9)
  node  │  22.17.0  │  package.json > engines > node (^22.17.0)

  Steps
  ──────────
  ▸ install
    $ pnpm install --frozen-lockfile --prefer-offline

  ▸ build
    $ pnpm run build

  Deploy
  ──────────
    $ caddy run --config /Caddyfile --adapter caddyfile 2>&1

docker buildx build web --build-arg BUILDKIT_SYNTAX=ghcr.io/railwayapp/railpack-frontend:latest -f web/railpack.json -t python-starter-template-javascript:51c797a294382f7975060c40fdd2da07edaae0de-dirty --platform linux/amd64 --cache-to type=inline --cache-from ghcr.io/iloveitaly/python-starter-template-javascript:latest --label org.opencontainers.image.description="Used for building javascript assets, not for deployment"
[+] Building 1.1s (10/19)                                                                                                                                                                        docker:orbstack
 => [internal] load build definition from railpack.json                                                                                                                                                     0.0s
 => => transferring dockerfile: 3.48kB                                                                                                                                                                      0.0s
 => resolve image config for docker-image://ghcr.io/railwayapp/railpack-frontend:latest                                                                                                                     0.6s
 => [auth] railwayapp/railpack-frontend:pull token for ghcr.io                                                                                                                                              0.0s
 => CACHED docker-image://ghcr.io/railwayapp/railpack-frontend:latest@sha256:46c992da3e89c3dd327506efd3b7fbe9507bb63d7343d00fd45ce23f1e8a532b                                                               0.0s
[...]

PR #226, opened 2025-08-20
fix: hard fail on invalid railpack.json

iloveitaly:
I noticed this while debugging my project's build failure. I kept getting a message about malformed `railpack.json` even though I did not provide one directly. It seems as through the generated `railpack.json` was incorrect but it did not cause the build to hard fail, it continued on without the generated `railpack.json`.

This change ensures that if invalid railpack.json is encountered, the CLI hard fails.

## Breaking Change

Previously, if JSON did not conform to the schema it would just be ignored and the build would continue. With this change, the build will halt and force the user to fix or discard their schema. I prefer this more explicit approach.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -1,7 +1,11 @@
 package core
 
 import (
+	"encoding/json"
+	"fmt"
 	"maps"
+	"os"
+	"path/filepath"
 	"slices"
 	"strings"
 
@@ -38,6 +42,26 @@ type BuildResult struct {
 	DetectedProviders []string                             `json:"detectedProviders,omitempty"`
 	Logs              []logger.Msg                         `json:"logs,omitempty"`
 	Success           bool                                 `json:"success,omitempty"`
+}
+
+func readConfigJSON(path string, v interface{}) error {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return err
+	}
+
+	jsonBytes, err := utils.StandardizeJSON([]byte(data))
+	if err != nil {
+		return err
+	}
+
+	stringData := string(jsonBytes)
+
+	if err := json.Unmarshal([]byte(stringData), v); err != nil {
+		return fmt.Errorf("error reading %s as JSON: %w", path, err)
+	}
+
+	return nil
 }
 
 func GenerateBuildPlan(app *app.App, env *app.Environment, options *GenerateBuildPlanOptions) *BuildResult {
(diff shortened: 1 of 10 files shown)


coffee-cup (commented):
What "generated railpack.json" file are you talking about? I kind of think that we should still be attempting to build if the users config file is malformed. Though could go either way

iloveitaly:
> What "generated railpack.json" file are you talking about?

This is hand-wavy on my end because I haven't read that part of the system.

What I was [seeing in my project](https://github.com/iloveitaly/python-starter-template/pull/4/files) (railpack build) was:

* no `railpack*` files
* I got a warning that there was an error reading a railpack.json
* I didn't really pay attention to it because (a) I didn't understand what the railpack.json was at the time (b) there wasn't a railpack.json file in my dir
* What is seems like is happening is there is a railpack.json 'generated' by the `build` command, but it's invalid, and then it's passed to the actual LLB logic where it's ignored
* This is the hand-wavy part since I decided to dig into the `.dockerignore` part of the system since I know that was causing issues.

I can report back when I dig into this more and understand exactly where this is failing.

iloveitaly:
> What "generated railpack.json" file are you talking about?

Ok, I discovered what was happening here:

In the build script I was testing, I had a `railpack prepare` call which dumped railpack.json. It seems like that command is resulting in invalid `railpack.json`. I was attempting to run `railpack build` directly, which I assumed would ignore `railpack.json` since I didn't specify it as a `--config-file` but it looks like it's read automatically.

Automatically consuming `railpack.json` without a warning feels wrong to me. I think we should only read this during `build` if specified explicitly. Wdyt?

> I kind of think that we should still be attempting to build if the users config file is malformed

I disagree here:

1. If a `railpack.json` exists it was either written by the user or `railpack prepare`. If it's invalid, it's probably a user error.
2. Right now there's only a warning that the schema is invalid and unless you are watching it really quickly it quickly gets gobbled up by a bunch of plan info printing and docker build progress.
3. If a dev breaks `railpack.json` accidentally and a build continues, it could cause a container to get deployed with a build configuration that was not intended, which could result in the wrong start script running, site to break, etc.

---

Two other issues I found here:

1. If an abs path is specified in `--config-file` it always fails since the file path is assumed to relative to the application source. This bit me since the build script I was using was passing an abs path.
2. If the passed config file DNE it is ignored, similar to a railpack.json that can't be parsed

I've updated the PR to correct these. My general thinking here is the provider/detection logic should be "magical" but any user-specified configuration inputs should be strict to guarantee deterministic build plans.

coffee-cup (commented):
> Automatically consuming railpack.json without a warning feels wrong to me. I think we should only read this during build if specified explicitly. Wdyt?

Can you expand on why it feels wrong? It is very standard across all tooling that supports a config file (Dockerfile, package.json, vite config, buildpacks).

Failing on an invalid `railpack.json` file makes sense then. Agree it will help avoid confusion. [In your case](https://github.com/railwayapp/railpack/issues/204#issuecomment-3086111605) I recommend naming the file `railpack-plan.json`

iloveitaly:
> Can you expand on why it feels wrong? It is very standard across all tooling that supports a config file (Dockerfile, package.json, vite config, buildpacks).

On second thought, I'm wrong here. Most build tooling _does_ automatically consume expected files, but most of the other CLI tools I use on a daily basis don't, which is why this felt wrong to me.

I think it's worth adding `path to config file to use (default: railpack.json)` to help output to make this more explicit.

coffee-cup:
> I think it's worth adding path to config file to use (default: railpack.json) to help output to make this more explicit.

That would make it much clearer!

iloveitaly:
@coffee-cup this one is ready for final review.

core/core.go
```diff
@@ -140,17 +164,27 @@ func GenerateConfigFromFile(app *app.App, env *app.Environment, options *Generat
 		configFileName = envConfigFileName
 	}
 
-	if !app.HasMatch(configFileName) {
+	// unless an absolute path, assume relative to the app source directory
+	var absConfigFileName string
+	if !filepath.IsAbs(configFileName) {
+		absConfigFileName = filepath.Join(app.Source, configFileName)
```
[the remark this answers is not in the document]
coffee-cup: The issue is that a platform (like Railway) may not want the user to be able to access files outside the app directory. We can't assume that the code being built also has access to all files on the host.

core/core.go
```diff
@@ -140,17 +164,23 @@ func GenerateConfigFromFile(app *app.App, env *app.Environment, options *Generat
 		configFileName = envConfigFileName
 	}
 
-	if !app.HasMatch(configFileName) {
+	// always assume config file path is relative to the app source directory
+	// https://github.com/railwayapp/railpack/pull/226
```
coffee-cup, line 168: I don't think we need to include a link to the issue/pr for every change like this

coffee-cup (approved):

core/core.go
theolujay: I see the point made with not leaving link to issue/pr for every change like this. I'd also like to say that I did find this particular one informative as I'm learning the codebase.

PR #237, opened 2025-09-01
fix: improved buildkit error message

iloveitaly:
The first-run error messages are pretty hard to understand for a new user who is running this locally.

We need a documentation section for this, but for now an improved error message walking the user through what do and where to look feels like a good first step.

### DIFF
--- a/buildkit/build.go
+++ b/buildkit/build.go
@@ -2,6 +2,7 @@ package buildkit
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


buildkit/build.go
```diff
@@ -47,12 +47,17 @@ func BuildWithBuildkitClient(appDir string, plan *plan.BuildPlan, opts BuildWith
 
 	buildkitHost := os.Getenv("BUILDKIT_HOST")
 	if buildkitHost == "" {
-		log.Error("BUILDKIT_HOST environment variable is not set")
-		return fmt.Errorf("BUILDKIT_HOST environment variable is not set")
+		return fmt.Errorf(`BUILDKIT_HOST environment variable is not set.
```
iloveitaly, line 50: multiline error messages do not seem like idiomatic go, but I don't see a better mechanism to do this?

Maybe better to use the `cxt.logger.LogError` here?

PR #238, opened 2025-09-01
test: assert against exact yarn package manager versions

iloveitaly:
While adding yarn 1 support, I noticed:

* the actual yarn version was not being tested in the test.json
* the wrong yarn version was being used on yarn 2 + 3

This PR fixes those two issues.


examples/node-yarn-2-node-linker/package.json
```diff
@@ -1,16 +1,12 @@
 {
   "name": "node-yarn-2-node-modules",
-  "packageManager": "yarn@4.7.0",
```
coffee-cup, line 3: Good catch!

PR #240, opened 2025-09-01
fix: fail early if no shell start script is detected, stop globbing shell script

iloveitaly:
I ran into this issue while working on standardizing local layer additions for docker ignore support.

* The start script is sent through a file globbing match. Fuzzy matching against the start script for a container seems like a bad idea. We should opt to be more explicit here.
* If a start script is not detected, but the shell provider is specified (either inferred or specified in the railpack.json) the user gets an obscure `chmod +x` during build, which is hard to read and understand

I have it on my TODO to look through other providers to ensure start scripts are not fuzzy matched.

**Question:** `app.HasMatch` is in a lot of places throughout the codebase where you would intuitively expect the input path _must_ correspond to a single, particular file. Is there an architectural reason why file glob matches are used instead?

### DIFF
--- a/core/app/app.go
+++ b/core/app/app.go
@@ -90,6 +90,14 @@ func (a *App) findGlob(pattern string) ([]string, error) {
 	return matches, nil
 }
 
+// Check if a relative file exists in the app's source directory
+func (a *App) HasFile(path string) bool {
+	fullPath := filepath.Join(a.Source, path)
+
+	_, err := os.Stat(fullPath)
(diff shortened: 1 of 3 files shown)


coffee-cup (approved):
> app.HasMatch is in a lot of places throughout the codebase where you would intuitively expect the input path must correspond to a single, particular file. Is there an architectural reason why file glob matches are used instead?

No particular reason besides preference for a smaller internal API

Issue #234, opened 2025-08-28
pyproject.toml with workspace dependencies fail


iloveitaly:
In this case, there's a workspace-local library (openapi spec generated client library) in pyproject.toml.

```
D venv ❯ railpack build .

╭────────────────╮
│ Railpack 0.3.0 │
╰────────────────╯

  ↳ Detected Python
  ↳ Using uv

  Packages
  ──────────
  python   │  3.11.13  │  railpack default (3.11)
  pipx     │  1.7.1    │  railpack default (latest)
  pipx:uv  │  0.8.13   │  railpack default (latest)

  Steps
  ──────────
  ▸ install
    $ uv sync --locked --no-dev --no-install-project
    $ uv sync --locked --no-dev --no-editable

  Deploy
  ──────────
    $ python main.py

[+] Building 1.3s (12/23)
 => CACHED docker-image://ghcr.io/railwayapp/railpack-runtime:latest                                                                                                                                                                                                                                                                             0.4s
 => => resolve ghcr.io/railwayapp/railpack-runtime:latest                                                                                                                                                                                                                                                                                        0.4s
 => loading .                                                                                                                                                                                                                                                                                                                                    0.4s
 => => transferring context: 1.28MB                                                                                                                                                                                                                                                                                                              0.4s
 => docker-image://ghcr.io/railwayapp/railpack-builder:latest                                                                                                                                                                                                                                                                                    0.4s
 => => resolve ghcr.io/railwayapp/railpack-builder:latest                                                                                                                                                                                                                                                                                        0.4s
 => CACHED install apt packages: python3-dev                                                                                                                                                                                                                                                                                                     0.0s
 => CACHED copy .tool-versions                                                                                                                                                                                                                                                                                                                   0.0s
 => CACHED mkdir /etc/mise                                                                                                                                                                                                                                                                                                                       0.0s
 => CACHED create mise config                                                                                                                                                                                                                                                                                                                    0.0s
 => CACHED install mise packages: pipx, pipx:uv, python                                                                                                                                                                                                                                                                                          0.0s
 => CACHED copy pyproject.toml                                                                                                                                                                                                                                                                                                                   0.0s
 => CACHED copy uv.lock                                                                                                                                                                                                                                                                                                                          0.0s
 => ERROR uv sync --locked --no-dev --no-install-project                                                                                                                                                                                                                                                                                         0.2s
 => CACHED copy /mise/shims                                                                                                                                                                                                                                                                                                                      0.0s
------
[...]

PR #241, opened 2025-09-01
refactor: use consistent local layers entrypoint

iloveitaly:
The main goal of this PR was to move all providers to avoid using `NewCopyCommand` and instead use `NewLocalLayer` so we have a single entrypoint for copying the filesystem into a container.

With this refactor in place, it'll make it much easier to update `NewLocalLayer` to parse the dockerfile and specify `Filters`.

This PR also bundles in a couple of other improvements:

* Support for workspace dependencies in uv. fixes https://github.com/railwayapp/railpack/issues/234
* `--show-plan` regression (binary output)
* Test for packaged pyproject.toml


core/providers/python/python.go
```diff
@@ -362,13 +368,33 @@ func (p *PythonProvider) copyInstallFiles(ctx *generate.GenerateContext, install
...
+	if err := ctx.App.ReadTOML("pyproject.toml", &pyproject); err == nil {
+		if tool, ok := pyproject["tool"].(map[string]any); ok {
+			if uv, ok := tool["uv"].(map[string]any); ok {
+				if _, exists := uv["workspace"]; exists {
+					return true
+				}
+			}
+		}
+	} else {
+		log.Infof("Failed to read pyproject.toml: %v", err)
```
coffee-cup, line 397: I think we should log with our custom logger so it gets saved to the plan

coffee-cup (approved):
I realize there is a potential caching difference with using the local layer input instead of a copy command since the order of the copying might be different (e.g. we run a command, then copy all local files into the step). But for the providers you changed, the order shouldn't change anything. Just bringing up in case you think otherwise.

iloveitaly:
> I realize there is a potential caching difference with using the local layer input instead of a copy command since the order of the copying might be different (e.g. we run a command, then copy all local files into the step)

At this point, I don't have a strong command over the caching system / how docker manages layer cache. I know it will change the layer cache, but my thinking is as long as we try to separate build & install phases (and only include the app source in `install`) it shouldn't negatively impact caching. Let me know if you see it differently.

PR #246, opened 2025-09-02
feat: support for expanded platform arguments

iloveitaly:
Confusing that `linux/arm64/v8` caused an error. Additionally, we can just use the parsing from `containerd` so we'll track with any platform additions in the future.

### DIFF
--- a/buildkit/platform.go
+++ b/buildkit/platform.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"runtime"
 
+	"github.com/containerd/platforms"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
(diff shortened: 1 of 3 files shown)


buildkit/platform.go
```diff
@@ -32,7 +33,27 @@ func DetermineBuildPlatformFromHost() BuildPlatform {
...
+	if platformStr == "" {
+		return DetermineBuildPlatformFromHost(), nil
+	}
+
+	platform, err := platforms.Parse(platformStr)
+	if err != nil {
+		return BuildPlatform{}, fmt.Errorf("invalid platform format: %s. Must be one of: linux/amd64, linux/arm64, etc", platformStr)
+	}
+
+	return BuildPlatform{
```
iloveitaly, line 46: We could probably just remove `BuildPlatform` completely and use the interface defined by `platforms`, but this fixes the immediate papercut without reworking the code more.

coffee-cup (approved):

PR #251, opened 2025-09-04
feat: support package.json engines for package manager and node versions

iloveitaly:
* Engines are now considered only if other version detection DNE, so lockfiles/configs take precedence over engines as a hint.
* This fixed a bug in the test suite where bun would be ignored
* Also fixed pnpm package manager semver extraction

### DIFF
--- a/.claude/commands/update_documentation.md
+++ b/.claude/commands/update_documentation.md
@@ -1,0 +1,3 @@
+Take a look at changes on this branch (compared to upstream/main) and update any relevant documentation in @docs/.
+
+Only update markdown documentation, do not make any code changes.


core/providers/node/package_manager.go
```diff
@@ -307,6 +324,11 @@ func (p PackageManager) GetPackageManagerPackages(ctx *generate.GenerateContext,
 	if p == PackageManagerBun {
 		bun := packages.Default("bun", "latest")
 
+		// Prefer explicit version from package.json engines over defaults
+		if packageJson != nil && packageJson.Engines != nil && packageJson.Engines["bun"] != "" {
+			packages.Version(bun, packageJson.Engines["bun"], "package.json > engines > bun")
+		}
```
coffee-cup: Nit: can this be extracted to something reusable since it is used in several place. Maybe something like

PR #257, opened 2025-09-10
test: use poetry cli to run python

iloveitaly:
Some users could modify the start command and reasonsable expect to use the `poetry` cli to run python, or other commands.

We ran into an issue where `poetry` was failing because it was using a copied version of python from an apt-install step which was removed and was now referencing invalid library files.

This ensures this specific case won't happen again.


iloveitaly:
@coffee-cup merging w/o review, lmk if you'd rather review this sort of thing.

Issue #255, opened 2025-09-09
Build fails if npm install in build command


coffee-cup:
If you have `npm ci` or `npm install` in the build command, then the build will fail. You can reproduce with

```
go run cmd/cli/main.go build examples/node-npm --build-cmd "npm ci"
```

and you get this error

```
 > npm ci:
0.236 npm warn config production Use `--omit=dev` instead.
0.425 npm error code EBUSY
0.425 npm error syscall rmdir
0.425 npm error path /app/node_modules/.cache
0.425 npm error errno -16
0.425 npm error EBUSY: resource busy or locked, rmdir '/app/node_modules/.cache'
0.426 npm error A complete log of this run can be found in: /root/.npm/_logs/2025-09-09T19_30_11_090Z-debug-0.log
------
 ERRO failed to solve: failed to solve: process "sh -c npm ci" did not complete successfully: exit code: 240
```

<img width="1155" height="1112" alt="Image" src="https://github.com/user-attachments/assets/e2813b37-da7c-4317-a2c6-c09b62f7ac7a" />

If you comment out [this line](https://github.com/railwayapp/railpack/blob/main/core/providers/node/node.go#L206), then the build works.

For some reason the `node_modules/.cache` during the build step is causing issues. I initially thought that adding this cache during the install step would help, but I still get the same error.

We might want to detect if npm install is being run in build and remove that cache?

PR #259, opened 2025-09-11
fix: allow node_modules removal in build phase

iloveitaly:
I really don't love the solution here, but it's not clear there is a better way:

* `npm ci` does not allow you to disable `node_modules` removal
* Users include `npm ci` in the `build` step, instead of the install step, which cause `node_modules` to attempt to be removed which then fails the build since it's a read-only mount when used as a cache directory.
* We *could* attempt to redirect various caches that use `node_modules/.cache` to something like `~/.node_modules_cache` but each tool uses their own way of configuring the cache path, and attempting to modify all of this tooling feels like it's not worth the squeeze (in addition to be confusing to users).
* We do need to document how to workaround "system" catches by defining your own `railpack.json`. I have that in my todo. This will be confusing to users.
* It seems like there is not a way around the read-only mountpath that buildkit uses for cache directories.
* Ideally the user could disable adding the `node_modules/.cache`, but I think it's better to _not_ create more knobs for users to tune and instead (a) try to do the right thing via ugly regex-style stuff like we've done in this PR and (b) enable the user to override everything via a `railpack.json` if they are outside the golden path.

What I decided on was "post processing" the build plan after it's been merged with the user-provided `railpack.json`. It's more complex, but if we attempted to omit `NODE_MODULES_CACHE` in the node provider:

* If a user supplied a `railpack.json` with `npm ci` they'd run into the same issue, which is odd
* `--build-cmd` also ran into the same issue

However, a very obvious case still fails:

```json
{
  "scripts": {
    "build": "npm ci"
  }
}
```

Without writing a web of complex code to analyize the content of `build` this would fail, and even that's a fools errand since they could shell out to anything that could attempt to `rm -rf node_modules`.

I still think the complexity here is worth it since it handles the common case and still gives advanced users the option to (a) understand `npm ci` should run in `install` and (b) write their own `railpack.json` to workaround the issue.

fixes https://github.com/railwayapp/railpack/issues/255

### DIFF
--- a/docs/src/content/docs/languages/node.md
+++ b/docs/src/content/docs/languages/node.md
@@ -169,3 +169,20 @@ Including:
 As well as a default cache for node modules:
 
 - Node modules: Caches `node_modules/.cache` (with the cache key `node-modules`)
+
+## Cache & Removing `node_modules`
+
+When you add custom build commands that remove `node_modules` (such as
+`npm ci`), Railpack automatically detects this and
+removes the `node_modules/.cache` directory from the cache configuration for
+those steps. This prevents `EBUSY: resource busy or locked` [errors](https://github.com/railwayapp/railpack/issues/255)
+that would otherwise occur when trying to remove a cached directory.
+
+This automatic handling applies to build steps that contain commands like:
+
+- `npm ci`
+- `rm -rf node_modules`
+- `rimraf node_modules`
+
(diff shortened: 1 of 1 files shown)


iloveitaly:
@coffee-cup this one is (finally) ready for final review!

PR #260, opened 2025-09-11
fix: use default mise backend for uv

iloveitaly:
The mise uv backend is asdf, but unlike poetry it is well supported and updated.

Long term, it's best for us to bet on the default backends for common pages as it will reduce our maintenance burden and compatibility issues.

I'd like to do this for everything but poetry, but I'm going to migrate one at a time in case there are any issues.


iloveitaly:
@coffee-cup sorry if I tagged you on this one, it's not ready yet!

iloveitaly:
Since this was already approved, and it's a small change, I'm going to merge without a review.

PR #261, opened 2025-09-11
feat: use py package versions from any mise-supported version specification file

iloveitaly:
Headline change here is pulling package versions determined by mise into the provider logic: this means `.tool-versions` and `mise.toml` language + runtime definitions used by the Python provider will override the `latest` defaults that are currently set. 

Notes:

* Python first because I know it best
* Limiting to python to (a) get this change out faster (b) reducing blast radius in case we break something
* `idiomatic_version_file_enable_tools` is not yet supported. This is on my TODO next, before rolling this out to more providers.
* The mise context interface seems odd to me. My Go knowledge is still pretty basic (had to dig into channels/concurrency for the first time for the test fix) so I could be missing what the best idiomatic Go structure is here.

## Multiple Expectations in `test.json`

I ran into a strange test suite bug that I had trouble getting to the bottom of: I had a multiline test case and the case was failing, even though the right content was specified:

* The integration tests assume that `expectedOutput` is a single line
* However, if you had multiple lines, `require.Contains` is asserted against the entire container stdout, so you'd get a failure with no detail

I've changed this:

* Support a single string or an array of strings in `expectedOutput` to make it easier to define multiple test cases
* Eliminate `require.Contains` on stdout and instead format the error string more explicitly


core/mise/mise.go
```diff
@@ -120,6 +121,55 @@ func (m *Mise) GetAllVersions(pkg, version string) ([]string, error) {
 	return versions, nil
 }
 
+// gets all package versions from mise that are defined in the app directory environment
+// this can include additional packages defined outside the app directory, but we filter those out
+func (m *Mise) GetPackageVersions(ctx MiseAppContext) (map[string]*MisePackageInfo, error) {
+	appDir := ctx.GetAppSource()
+	output, err := m.runCmd("--cd", appDir, "list", "--current", "--json")
```
coffee-cup, line 128: I think we should run in [paranoid](https://mise.jdx.dev/paranoid.html) mode. I am a bit worried about executing any user code here since that is by definition untrusted.

core/generate/mise_step_builder.go
```diff
@@ -17,6 +19,26 @@ const (
 	MiseInstallCommand  = "sh -c 'mise trust -a && mise install'"
 )
 
+// MiseListSource represents the source of a mise tool installation
+type MiseListSource struct {
+	Type string `json:"type"`
+	Path string `json:"path"`
+}
+
+// MiseListTool represents a tool in the mise list output
```
iloveitaly, line 28:
```suggestion
// output of `mise list`
```

iloveitaly:
@coffee-cup this is ready for a final review:

* Provider only needs to pass a list of packages to the mise step builder. Much cleaner, great suggestion.
* Refactored the interfaces + logic to be cleaner

Future improvements:

1. Add mise overrides to all providers
2. Test to ensure multiple mise configs in a single dir are respected (`.config/mise.toml`)
3. 1. Install all packages defined by mise configuration

core/providers/python/python.go
```diff
@@ -314,26 +314,36 @@ func (p *PythonProvider) InstallMisePackages(ctx *generate.GenerateContext, mise
...
 		miseStep.Default("pipx:uv", "latest")
+		packages = append(packages, "pipx:uv")
 	}
 
 	if p.hasPipfile(ctx) {
 		miseStep.Default("pipx:pipenv", "latest")
+		packages = append(packages, "pipx:pipenv")
 	}
 
+	miseStep.UseMiseVersions(ctx, packages)
```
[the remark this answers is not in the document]
iloveitaly: This would be cleaner right now, but I don't think we should use pipx for _everything_ (installing it with uv doesn't make any sense since they have a great native installer).

I'll clean this up in https://github.com/railwayapp/railpack/pull/260.

PR #263, opened 2025-09-12
feat: dockerignore support

iloveitaly:
Most of the heavy lifting was in the other PRs we've merged, so this one is pretty straightforward.

There's some follow-on work around removing the node_modules log, adding docs, adding this to meta output,  etc. I have that on my TODO.

### DIFF
--- a/core/plan/dockerignore_test.go
+++ b/core/plan/dockerignore_test.go
@@ -5,12 +5,20 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/railwayapp/railpack/core/app"
 	"github.com/stretchr/testify/require"
 )
 
 func TestCheckAndParseDockerignore(t *testing.T) {
 	t.Run("nonexistent dockerignore", func(t *testing.T) {
-		excludes, includes, err := CheckAndParseDockerignore("/nonexistent/path")
+		tempDir, err := os.MkdirTemp("", "dockerignore-test")
+		require.NoError(t, err)
+		defer os.RemoveAll(tempDir)
+
+		testApp, err := app.NewApp(tempDir)
+		require.NoError(t, err)
+
+		excludes, includes, err := CheckAndParseDockerignore(testApp)
 		require.NoError(t, err)
 		require.Nil(t, excludes)
 		require.Nil(t, includes)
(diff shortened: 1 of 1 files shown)


core/plan/dockerignore.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/moby/patternmatcher/ignorefile"
+)
+
+// checks if a .dockerignore file exists in the app directory and parses it
+func CheckAndParseDockerignore(repoPath string) ([]string, []string, error) {
```
coffee-cup, line 12: Thoughts on passing in the app directly here? All filesystem reads up until now (should) go through the app abstraction and keeping that in place has future benefits IMO (we could read from a GitHub repo using the API or directly from a compressed directory). Far out, but if we are only reading the file once in this function, it is simple enough to keep that abstraction in place.

CLAUDE.md
```diff
@@ -22,9 +17,11 @@ with support for Node, Python, Go, PHP, and more.
 
 # Bash commands
 
-- `mise run build` - Build the CLI binary
+- `mise run build` - Build the CLI binary (`./bin/railpack`), use this instead of `go build` (`./bin/railpack`), use this instead of `go build`
```
coffee-cup, line 20:
```suggestion
- `mise run build` - Build the CLI binary (`./bin/railpack`), use this instead of `go build`
```

CLAUDE.md
```diff
@@ -22,9 +17,11 @@ with support for Node, Python, Go, PHP, and more.
...
 - `mise run check` - Run linting, formatting, and static analysis
 - `mise run test` - Run unit tests
+- `mise run test-integration` - Runs *all* integration tests. Never run this. Instead, only run the test that you are currently working on, e.g. `mise run test-integration -- -run "TestExamplesIntegration/config-file"`
+- `mise run test-integration` - Runs *all* integration tests. Never run this. Instead, only run the test that you are currently working on, e.g. `mise run test-integration -- -run "TestExamplesIntegration/config-file"`
```
coffee-cup, line 24:
```suggestion
```

coffee-cup (approved):
Looks good! Just one comment about the app abstraction but feel free to ignore it.

This will be a huge improvement when working with Railpack 😃

CLAUDE.md
iloveitaly: thanks for catching these

core/plan/dockerignore.go
iloveitaly: good call, just refactored this

Issue #165, opened 2025-06-11
Maven multi-module support


derkoe:
Most Java applications are built with multiple modules where one module is the actual application to run.

Here is an example: https://github.com/porscheinformatik/angular-spring-heroes - here the application is under `heroes-webapp`.

Railpack should support this pattern. This could be achieved by finding the executable application automatically or via a config variable.

iloveitaly:
@derkoe Thanks for the report here and sorry for the delay in getting back to you!

I'm not too familiar with Java: is there any chance there's a simpler repo that I can use to add an [integration test](https://github.com/railwayapp/railpack/tree/main/examples) for this, which would make it easier for me to add support for this case in railpack?

derkoe:
@iloveitaly added an example in #265 but I have no clue how to implement that in Railpack

iloveitaly:
@derkoe amazing, thanks so much! I should be able to spend some time on it this week to see if I can build out support.

PR #265, opened 2025-09-15
feat: add Java multimodule example

derkoe:
Adds an example for a Maven multi-module project - this will fail with the current Railpack implementation since it does not support Maven multi-module projects.

See #165


iloveitaly:
This was a really helpful example. I've pulled this into #330 (nearly done!)

PR #278, opened 2025-09-25
feat: allow container variables to be specified

iloveitaly:
Issue I ran into when attempting to replace nixpacks with railpack in my template project:

* `--env` currently represents a secret, not an [environment variable as specified here](https://railpack.com/architecture/secrets/#_top)
* There's not a way to define a environment variable when running `railpack build` which makes it challenging to reproduce a production railway build locally.

Here's what I propose:

* Breaking change: move current `--env` functionality to `--secret`
* Add a `--env` flag which works like the `--var` flag proposed in this PR

This code is a rough draft; I need to give it another review. I wanted to get your take on this before spending more time on it.

### DIFF
--- a/cli/build.go
+++ b/cli/build.go
@@ -133,7 +133,7 @@ var BuildCommand = &cli.Command{
 func validateSecrets(plan *plan.BuildPlan, env *app.Environment) error {
 	for _, secret := range plan.Secrets {
 		if _, ok := env.Variables[secret]; !ok {
-			return fmt.Errorf("missing environment variable: %s. Please set using --env %s=%s", secret, secret, "...")
+			return fmt.Errorf("missing environment variable: %s. Please set using --secret %s=%s", secret, secret, "...")
 		}
 	}
 	return nil
--- a/cli/common.go
+++ b/cli/common.go
@@ -23,6 +23,10 @@ func commonPlanFlags() []cli.Flag {
 			Usage:   "environment variables to set",
 		},
 		&cli.StringSliceFlag{
+			Name:  "var",
+			Usage: "variables to set",
+		},
+		&cli.StringSliceFlag{
 			Name:  "previous",
 			Usage: "versions of packages used for previous builds (e.g. 'package@version')",
 		},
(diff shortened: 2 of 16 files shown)


coffee-cup (commented):
Curious why you need a way of specifying envs vs secrets? How are they handled differently? Having both is unnecessary complexity IMO and I'd love if we could just have a single way of handling envs/secrets (and all user provided "envs" are considered secret)

iloveitaly:
@coffee-cup 

* There's a set of variables that are NOT secrets, but have to be set on the container. I want to set them on the container to remove additional config within my hosting environment.
* Examples: [python and timezone config](https://github.com/iloveitaly/python-starter-template/blob/6014475e8015df5923e00da87915bb1904811dc6/env/all.sh#L3-L17), environment config (`PYTHON_ENV=production`, which is a non-standard env var), [build metadata](https://github.com/iloveitaly/python-starter-template/blob/6014475e8015df5923e00da87915bb1904811dc6/just/build.just#L17-L19)
  * I realize some of this (build metadata) is handled by railway, but I'd prefer to be able to inject + control this build metadata on my end so I run the application with the same production-like configuration locally. I realize this is a bit idiosyncratic :)

iloveitaly:
Another use-case here:

Right now, there is not a way to override the provider-configured variables (such as `PYTHONUNBUFFERED=`) in a way that _all steps_ use the overridden values. Again, probably a niche need but I'd like to avoid the case where a more advanced user has to fallback to a dockerfile.

Additionally, the implementation I have in this PR does not enable the user to override provider-supplied variables.

Need to put some more thought into the best solution here and if it's worth doing anything, but wanted to drop this thought here.

coffee-cup:
Very fair use cases. I just worry about complicating the API with two ways of adding environment variables. This also doesn't translate well to platforms that use Railpack (e.g. Railway). How would users set these variables?

iloveitaly:
Yeah I hear you and share your concerns. This is low-ish pri for me right now ahead of (a) getting all providers to respect mise versions and (b) squash all of the user bugs, will put more effort here later on.

iloveitaly:
@coffee-cup I put some thought into this. Here's what I think we should do:

1. We should use `env` instead of `variables` in `railpack.json`. Variables is a weird name for environment variables:
    * GHA uses `env` for their map of environment variables
    * Shells use `env` to dump all environment variables
    * `env` is used for all of the most popular languages `os.getenv`, `ENV.fetch`, `process.env`
2. Change `--env` => `--secret` CLI option and use `--env`
3. Add a top-level `railpack.json` option for an `env` key
4. To your previous question, Railway users can use `railpack.json` to customize environment variables

This would introduce breaking changes, but I think it's worth it to get the nomenclature right to eliminate confusion in the future (plus... we are in < 1.0 land still).

I've outlined how this would work in the updated `secrets.md` on this PR. I have not updated the code on this PR to reflect the changes in `secrets.md`. 

What do you think?

coffee-cup:
These changes make sense. Some concerns
- We need an incremental way of rolling this out and deprecating the current `--env` behaviour. Likely just support both and show a warning and then later on remove the old behaviour (Railway can switch to the new flags at this time too)
- The "Custom Frontend" example needs to handle both envs and secrets too right? like the prepare/build with buildkit format. I think it should be simple to support, I just noticed it missing in this PR

PR #285, opened 2025-09-27
fix: generated railpack.json

iloveitaly:
The PR fixes a couple of issues:

* The generated JSON schema is incorrect (requires provider when it is not required)
* Some of the railpack.json
* `prepare` generates railpack.json which (a) cannot be parsed by prepare/build and (b) is invalid according to the schema

TODO

- [ ] submit railpack.json to schema registry and see if we can list as JSONC
- [ ] docs for json schema hint in repo for jsonc/json5 support
- [ ] should generate railpack.json using `prepare` and check it against generated schema instead of just testing against customized railpack.json


iloveitaly:
railpack is in the global schema config repo. The other changes proposed here are misunderstood.

I've [updated the documentation](https://github.com/railwayapp/railpack/pull/662) to make it easier to understand what is happening with `railpack.json`.

PR #290, opened 2025-10-07
docs: update python with mise-compatible version files

crisog:
Now specifies mise-compatible version files such as:
- `.python-version`
- `.tool-versions`
- `mise.toml`


iloveitaly:
Thanks for this!

PR #293, opened 2025-10-12
fix: fit variable name width

thesamgordon:
I understand that this is a pretty hacky approach but it seems to work at all resolutions. It looks very messy without this change.

Old:
image

New:
image


iloveitaly:
This was a great catch! I was able to fix it with a CSS style so we don't need to worry about this on other pages.

PR #300, opened 2025-10-18
feat: respect go version specified by mise

iloveitaly:
Similar to what we recently launched for python.

Since the project is written in go, this required the mise isolation work that was recently merged to work well with our test suite.


coffee-cup (approved):
Looks good pending tests

PR #301, opened 2025-10-21
fix: disable Python compilation to avoid package incompatibility issues

coffee-cup:
Disables Python compilation by setting MISE_PYTHON_COMPILE=false to avoid incompatibility issues with some packages.


iloveitaly:
@coffee-cup curious which packages were erroring out on this?

coffee-cup:
@iloveitaly A user had an issue when they deployed right as Python 3.11.14 became available on Mise but there were no precompiled binaries. For whatever reason the compiled Python version did not have some libs available (like sqlite) and caused runtime crashes.

PR #304, opened 2025-10-21
fix: use channel key not version from rust-toolchain.toml

### LINKED ISSUE #296: Incorrect `rust-toolchain.toml` support
I was adding a new Rust project and noticed it doesn't inherit my Rust version correctly. I use `rust-toolchain.toml`, like so:

```toml
[toolchain]
profile = "default"
channel = "1.90.0"
```

The Railpack docs (https://railpack.com/languages/rust#versions) mention that only `toolchain.version` is supported, and not `toolchain.channel`, but the former is not a valid field: https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

Are we expected to deviate from conventions here?

iloveitaly:
fixes #296


iloveitaly:
@coffee-cup I think this one is fine to merge w/o review. Let me know if you disagree.

PR #305, opened 2025-10-21
refactor: cleanup rust provider

iloveitaly:
The way the rust compilation logic was structured seemed odd and made it harder to grok what was going on. This cleans it up and keeps the logic out of bash scripts.

Requesting a review to make sure I'm not missing some unintuitive reason why it was structured this way.


coffee-cup (approved):
Much cleaner!

PR #306, opened 2025-10-22
feat: pull elixir versions from mise config

iloveitaly:
* Pull elixir and erlang versions from mise configuration
* Better docker compose cleanup (causing issues locally)
* Run full-stack elixir projects, including database config

TODO

- [x] wait until docker-compose support is merged so we can fix elixir builds to actually run and include version information in the output


iloveitaly:
@coffee-cup merging this w/o review—let me know if you'd like to do it differently!

PR #318, opened 2025-10-30
feat(gleam): add GLEAM_INCLUDE_SOURCE option

aleksrutins:
This allows you to include the source tree in the final container, useful for things like [lustre-ui](https://github.com/lustre-labs/ui). I don't see a way to configure an automated test with `RAILPACK_` environment variables, but I did test this manually. The `gleam-include-source` test currently tests that, if that variable does what it says it does, it should work, but it doesn't test whether or not the variable does what it should.


iloveitaly:
I had to verify, but you can set ENV vars for tests. I've updated this PR to do this!

PR #319, opened 2025-10-30
feat: C/C++ providers

### LINKED ISSUE #164: C++ support
What would be involved in adding support for CPP?

I would love to be able to use this to build ROS2 projects (which are typically a combination of CPP and python packages)

aleksrutins:
Closes #164 

This PR adds a few providers:
- [x] ROS
- [x] CMake/Meson

### DIFF
--- a/core/__snapshots__/TestGenerateBuildPlanForExamples_python-latest-psycopg_1.snap.json
+++ b/core/__snapshots__/TestGenerateBuildPlanForExamples_python-latest-psycopg_1.snap.json
@@ -111,7 +111,8 @@
     "MISE_NODE_VERIFY": "false",
     "MISE_PARANOID": "1",
     "MISE_PYTHON_COMPILE": "false",
-    "MISE_SHIMS_DIR": "/mise/shims"
+    "MISE_SHIMS_DIR": "/mise/shims",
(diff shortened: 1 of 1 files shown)


aleksrutins:
Based on #164 and looking around based on that, it seems that Docker images are a relatively common way to package ROS packages, and normally that requires a decent bit of setup. As far as the other providers, I'd use the CMake provider. The Meson provider I'd probably also use, just not as often. They should all be relatively easy to keep up to date - the CMake and Meson providers default to just the latest versions, and so shouldn't require much maintenance at all, and all the ROS provider should need is an occasional base image update.

As far as merging the providers, I'd be willing to do that for the Meson and CMake providers - the ROS provider, though, is a completely different beast, so I'd rather keep it separate.

iloveitaly:
@aleksrutins thanks for your amazing work here and sorry for the delay!

We decided not to support ROS. I'd never heard of it before—looks like a super cool framework—but I don't see a way any cloud provider is going host a ROS project. We should make it more clear that Railpack is focused on building containers for cloud deployment and not on-device images.

The C++ provider looks great. If you can remove the ROS provider + docs, I can get the C++ support merged.

aleksrutins:
Weirdly enough, the snapshots succeed on my machine.

PR #322, opened 2025-11-06
docs: add explanation for MISE_PYTHON_COMPILE=false

iloveitaly:
comment on why we are disabling mise py compilation


iloveitaly:
claude really failed out on this one

PR #324, opened 2025-11-06
feat: set pipx.uvx = true when pipx backend is used in Python provider

iloveitaly:
Add support for mise settings configuration and set pipx.uvx = true when using pipx in the Python provider. This configures pipx to use uvx as the backend, as documented at:

https://mise.jdx.dev/dev-tools/backends/pipx.html

This isn't zero risk, but I'd rather use uv everywhere we can (the tooling is just so good):

> This should just mean that it installs much faster, but see below to disable or configure since occasionally tools don't work with uvx.

---

This also lays the groundwork for adding settings config which can be used to enable paranoid mode in the future.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -25,6 +25,9 @@ with support for Node, Python, Go, PHP, and more.
 - Seriously, do not write comments that are obvious from the code itself.
 - Do not write one-line functions
 - Always use the App abstraction for file system operations.
+- When configuring Mise settings, prefer environment variables (e.g.
+  `MISE_PIPX_UVX`) over TOML settings to maintain consistency with existing
+  patterns in the codebase.
 
 # Workflow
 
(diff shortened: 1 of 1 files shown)


core/providers/python/python.go
```diff
@@ -321,6 +321,9 @@ func (p *PythonProvider) InstallMisePackages(ctx *generate.GenerateContext, mise
 	if p.hasPoetry(ctx) || p.hasPdm(ctx) || p.hasPipfile(ctx) {
 		miseStep.Default("pipx", "latest")
 		packages = append(packages, "pipx")
+
+		// prefer to use uv tooling as much as we can
+		miseStep.Settings["pipx.uvx"] = true
```
coffee-cup: All Mise settings can be configured through an environment variable, can we do that instead of running separate mise commands (we already do this in a few places)? For this we could set `MISE_PIPX_UVX=true` (which I think it is by default).

iloveitaly:
> All Mise settings can be configured through an environment variable, can we do that

Nice, this is better.

> which I think it is by default

it's only default if `uv` is installed, which is not true in the case we when are using pipx.

PR #329, opened 2025-11-06
chore: update mise

iloveitaly:
after mise version `2025.10.13` there was some change to the mise trusted dir setup. It looks like it was undocumented change.

* We specify both a specific directory to trust when running on the host and a parent dir to STOP searching at. This ensures that only mise config within the repo is used.
* Also isolated trust state to a /tmp dir
* It was easier to pull in the GHA changes so we use mise to configure binaries on CI than to fix the mise stuff on the existing GHA setup.
* `MISE_NO_CONFIG=1` for any version querying to avoid trust issues
* `MISE_SYSTEM_DIR` on host mise execution to avoid picking up global mise config

### DIFF
--- a/core/mise/install.go
+++ b/core/mise/install.go
@@ -17,7 +17,7 @@ import (
 )
 
 const (
-	miseVersion       = "2025.11.2"
+	miseVersion       = "2025.11.3"
 	githubReleaseBase = "https://github.com/jdx/mise/releases/download"
 )
 


iloveitaly:
@coffee-cup Requesting a quick review here, there are some pretty significant changes to the Mize side of things, so I would like a quick check to make sure I'm not missing anything.

PR #330, opened 2025-11-07
feat: multimodule maven support

iloveitaly:
Fixes https://github.com/railwayapp/railpack/issues/165

### DIFF
--- a/docs/src/content/docs/languages/java.md
+++ b/docs/src/content/docs/languages/java.md
@@ -51,18 +51,13 @@ If JAR selection fails, specify a custom start command.
 
 ## Custom Start Command
 
-To override the default start command, create a `railpack.toml` file in
+To override the default start command, create a `railpack.json` file in
 your project root:
 
-```toml
-[build]
-# Custom command to run your application
-startCommand = "java -jar server/target/my-app.jar"
-```
-
-You can also use shell commands to dynamically find JARs:
-
-```toml
-[build]
-startCommand = "java -jar $(find . -name 'my-app-*.jar' | head -1)"
+```json
+{
+  "deploy": {
+    "startCommand": "java -jar server/target/my-app.jar"
+  }
+}
 ```


docs/src/content/docs/languages/java.md
```diff
@@ -27,3 +27,42 @@ The Java version is determined in the following order:
...
+
+- Your project has multiple executable JARs
+- JARs don't follow standard naming conventions
+- The build produces only thin JARs (without bundled dependencies)
+
+If JAR selection fails, specify a custom start command.
+
+## Custom Start Command
+
+To override the default start command, create a `railpack.toml` file in
```
iloveitaly: @claude Where did you come up with `railpack.toml`? This config has to be in `railpack.json`.

docs/src/content/docs/languages/java.md
```diff
@@ -27,3 +27,37 @@ The Java version is determined in the following order:
 | ------------------------- | --------------------------- | ------- |
 | `RAILPACK_JDK_VERSION`    | Override the JDK version    | `17`    |
 | `RAILPACK_GRADLE_VERSION` | Override the Gradle version | `8.5`   |
+
+## Multi-Module Projects
```
iloveitaly: @derkoe I don't have a lot of experience with Java. I'm curious what you think of this approach and if this makes sense for your particular use case.

iloveitaly:
@coffee-cup This PR was very much LLM-inspired, as I have very little understanding or experience with Java. I'm not sure if you have more experience than I, but it would be great to get a second pair of eyes on this to make sure I'm not missing something obvious.

coffee-cup (approved):
I am absolutely not the person to ask for Java stuff either :/
I would say good to merge as long as we have tests

docs/src/content/docs/languages/java.md
derkoe: @iloveitaly I've tried it and I think you will have to redo the whole pack implementation for Java. 

I've tried it on https://github.com/porscheinformatik/angular-spring-heroes/ and when you look in the container image I can see the following ending in there:
- the whole .git folder (aka the whole repository)
- the whole source code for the application
- node_modules for the whole app (even though not used during runtime)

Here is a screenshot from dive:
grafik

The built app ends up in `/app/heroes-webapp/target/heroes-webapp-1.0.0.BUILD-SNAPSHOT.jar` but this is not found by the launcher:

```
❯ docker run -it angular-spring-heroes
Error: -jar requires jar file specification
```

PR #344, opened 2025-11-12
feat: mise idiomatic file parsing

iloveitaly:
This eliminates the need for us to maintain parsers for various languages and gives us support for free for a bunch of additional idiomatic files that we don't currently support


iloveitaly:
@coffee-cup this feels ok to self-merge given the scope of the change is small, but it will pick up additional version files, which will change behavior for some users.

In a separate PR I'll rip out all of the parsers we have that we don't need anymore.

PR #347, opened 2025-11-13
refactor: replace HasMatch with HasFile for specific file paths

iloveitaly:
Replace fuzzy matching (HasMatch) with exact file path checks (HasFile) in all providers where specific files are being checked. This prevents false positives from glob pattern matching when checking for specific file names.

Directory checks (e.g. src/bin, node_modules, public, script) continue to use HasMatch as they need pattern/directory matching.

### DIFF
--- a/core/generate/mise_step_builder.go
+++ b/core/generate/mise_step_builder.go
@@ -325,7 +325,7 @@ func (b *MiseStepBuilder) GetSupportingMiseConfigFiles(path string) []string {
 	files := []string{}
 
 	for _, file := range miseConfigFiles {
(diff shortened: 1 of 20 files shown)


iloveitaly:
@coffee-cup this could introduce some subtle changes to provider detection that will be hard to diagnose with users. I do think it's best to get this change in place now even if there's some risk.

Matching against explicit files as opposed to fuzzy matching will make the provivder detection more deterministic and when it _doesn't_ work for a user, we can find out about it and then add an integration test for that specific case and make sure it is properly handled.

PR #354, opened 2025-11-13
test: add YJIT assertion to ruby-vanilla example

iloveitaly:
This test currently fails because YJIT is not automatically enabled by Railpack for Ruby 3.2+ apps. The test expects "YJIT: enabled" but will output "YJIT: disabled" until we add RUBY_YJIT_ENABLE=1 to the Ruby environment variables.

Related to performance audit task.


iloveitaly:
`examples/ruby-3/` has a yjit test, closing this out.

PR #356, opened 2025-11-17
fix: require GPG verification for all but the recent node versions

iloveitaly:


iloveitaly:
Closing this out and instead recommending configuration within the build plan logging output.

PR #363, opened 2025-11-17
fix: exclude venv if no dockerignore is present

iloveitaly:


iloveitaly:
closing in favor of https://github.com/railwayapp/railpack/pull/373

PR #365, opened 2025-11-18
feat: enable GPG signature verification for mise installs

iloveitaly:
* Adds MISE_GPG_VERIFY=1 environment variable to enable GPG signature verification when installing tools via mise in container builds. This has a `None` default.
* It's unclear what the blast impact of this is. All of the test fail, but this could cause additional failures in production (due to the latest node verification issue we ran into in the past). However, I think this is worth the risk. We should opt for additional security measures when we can.


coffee-cup (approved):
Thoughts on having this disabled by default initially and opt-in? I'm just worried that we start breaking a bunch of builds in production and then just disable it (and then what is the point?)

iloveitaly:
in the interest of making more builds succeed by default, I'm added a recommendation entry for this and closing this PR out. May add a build config suggestion for this as well.

PR #371, opened 2025-11-18
test: oldest supported node version

iloveitaly:
Adds a minimal example using Node v15 via mise.toml to verify railpack can build projects with older Node versions.


iloveitaly:
@coffee-cup we didn't have tests for the newest node version, which is what caused the issue today. But, we also didn't have a test for the oldest. 

As part of this, I also added a LTS policy to the docs, which is what I'd like your eyes on.

PR #375, opened 2025-11-22
feat: add python-oldest and python-latest version examples

iloveitaly:
Add two new Python examples to test version range support:
- python-oldest: tests Python 3.10 (oldest supported LTS version)
- python-latest: tests latest available Python version via mise

Also update Python documentation to indicate supported version range (3.10+), matching Python's official support policy.


iloveitaly:
@coffee-cup FYI documenting the same support window for python as well.

PR #386, opened 2025-12-09
chore: use `bun.lock` instead of `bun.lockb` in bun example

JamBalaya56562:
## Summary

Update the node-bun example configuration and lockfile to align with newer Bun and TypeScript settings.

Enhancements:
- Adjust TypeScript compiler options in the node-bun example for ESNext modules with preserved module syntax and stricter type-checking flags.
- Relax the TypeScript peer dependency range in the node-bun example to allow any ^5.x release.
- Add a Bun lockfile to the node-bun example to standardize dependency resolution.

Documentation:
- Refresh the node-bun example README with the current Bun version and documentation URL.


iloveitaly:
Thank you for the contribution! There are already some other projects with `bun.lock`, so I'd keep to keep this example using the older `lockb` format to ensure this case is tested.

I should document this in the repo to make that explicit.

PR #412, opened 2026-01-03
fix: remove duplicate bun version file read

### LINKED ISSUE #413: fix: remove duplicate bun version file read
Remove duplicate `.bun-version` file read in Node provider. The file was being read twice with identical logic.

Relates to #412

luojiyin1987:
Fixes #413 
## Summary
- Remove duplicate `.bun-version` file read in Node provider
- The file was being read twice with identical logic

## Test plan
- Existing integration tests cover this code path


iloveitaly:
Nice find! Thanks for the fix.

PR #414, opened 2026-01-03
feat: use consistent node version detection for bun

luojiyin1987:
## Summary

- Use consistent Node version detection for Bun projects
- Ensures the `.nvmrc` or `package.json` `engines.node` version is used when available

## Test plan

- [x] Unit tests pass
- [x] Integration tests pass


iloveitaly:
thanks for your contribution here! I have a PR for this that I've been working on here https://github.com/railwayapp/railpack/pull/359 that I'm going to use instead because of some other changes I'm bundling in.

PR #435, opened 2026-01-17
fix: .dockerignore negations causing BuildKit lstat failures

Reidaa:
I was trying to use Railpack on some obscure side project, and writing my own Dockerfile seemed so uncivilised compared to using Railpack.
I kept running into `ERROR: failed to build: failed to solve: lstat /.vscode/tasks.json: no such file or directory` during the build phase. I had neither created nor pushed any file named `tasks.json` to the project’s repository, so why would it be looking for it?
Some quick investigation led to the discovery that the pattern `!.vscode/tasks.json` in my `.dockerignore` was being wrongly interpreted by Railpack. 
I unleashed Codex at it and here we are. 

## Problem
Negated `.dockerignore` entries (e.g. `!/.vscode/tasks.json`) were being converted into explicit include paths in the generated plan. That could cause BuildKit to try to `lstat` absolute host paths like `/.vscode/tasks.json`, failing when those files don’t exist in the build context.

## Changes
- Preserve parsed `.dockerignore` patterns as-is (including `!` negations) in `DockerignoreContext`.
- Apply `.dockerignore` patterns only as local-layer exclude patterns while keeping local includes as `["."]` to avoid emitting explicit include copies.
- Add/update tests to cover negated dockerignore entries.
## Testing

- `mise run check`
- `mise run test`
- `mise run test-integration-cwd` (from `examples/dockerignore/`) ✅

### DIFF
--- a/core/generate/context.go
+++ b/core/generate/context.go
@@ -71,14 +71,13 @@ func NewGenerateContext(app *a.App, env *a.Environment, config *config.Config, l
 	}
 
 	dockerignoreCtx := plan.NewDockerignoreContext(app)
-	excludes, includes, err := dockerignoreCtx.ParseWithLogging(logger)
+	patterns, err := dockerignoreCtx.ParseWithLogging(logger)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse .dockerignore: %w", err)
 	}
 
-	if excludes != nil || includes != nil {
-		log.Debugf("Dockerignore exclude patterns: %v", excludes)
-		log.Debugf("Dockerignore include patterns: %v", includes)
+	if patterns != nil {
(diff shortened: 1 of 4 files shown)


iloveitaly:
> I was trying to use Railpack on some obscure side project, and writing my own Dockerfile seemed so uncivilised compared to using Railpack.

YES! This is exactly what we are trying to support with this project :) 

> kept running into `ERROR: failed to build: failed to solve: lstat /.vscode/tasks.json: no such file or directory` during the build phase. I had neither created nor pushed any file named `tasks.json` to the project’s repository, so why would it be looking for it?

Great find! This PR was a great starting point, but I wanted to add some more detailed tests for this case so I pulled it into a separate PR: https://github.com/railwayapp/railpack/pull/437

Closing this one out.

PR #440, opened 2026-01-17
docs: do not mark procfile as deprecated

iloveitaly:
This is definitely a personal preference, but I really like Procfiles.

* They are not vendor specific
* They self-document what the production start commands are in-code
* They can be run locally using foreman, honcho, etc to make it easy to run systems *exactly* as they are in production

I don't think we should mark them as deprecated.


iloveitaly:
IRL recap: `railpack.json` config is better, but we will support procfiles long into the future, but should not recommend them strongly.

PR #450, opened 2026-01-22
fix(python): detect production-only deps

iloveitaly:
this is just a draft, need to think on this further, but the current detection methods (both here and in node) are problematic.


iloveitaly:
this was added in the recent playwright python support

Issue #455, opened 2026-02-08
Bug: declaring `secrets` in config file removes all other environment variables from build steps


VBsec:
## Description

Adding a `secrets` field to `railpack.json` causes all environment variables **not** listed in `secrets` to be completely unavailable to build steps. This contradicts the [documentation](https://railpack.com/architecture/secrets) which states that secrets and environment variables are treated separately:

> Build secrets and environment variables are treated separately. The main differences being:
> - Environment variables are saved in the final image and should not contain sensitive information. Since they are in the final image, providers can add variables that will be available to the app at runtime.
> - Secrets are never logged or saved in the build logs. They are also only available at build time and not saved to the final image.

## Expected behavior

The `secrets` field should only control which secret names are part of the build plan for cache invalidation purposes. It should not affect the availability of other environment variables passed via `--env`.

## Actual behavior

The `secrets` field acts as an exclusive whitelist — only the env vars named in `secrets` are mounted via `llb.AddSecret` in BuildKit. All other env vars passed via `--env` are silently dropped from every build step.

## Reproduction

Given a `railpack.json` **without** a `secrets` field:

```json
{
  "steps": {
    "build": {
      "commands": ["...", "./check-env.sh"]
    }
  },
  "deploy": { "startCommand": "node main.js" }
}
```

```bash
railpack build \
  --env DATABASE_URL=postgres://localhost/mydb \
  --env SENTRY_AUTH_TOKEN=sntrx_abc \
  --env VITE_APP_TITLE=MyApp \
  --env MY_SECRET=s3cret .
```

Build step output — all 4 vars available:

```
DATABASE_URL=postgres://localhost/mydb
SENTRY_AUTH_TOKEN=sntrx_abc
VITE_APP_TITLE=MyApp
MY_SECRET=s3cret
```

Now add `"secrets": ["VITE_APP_TITLE"]` to the same config:

```json
{
  "secrets": ["VITE_APP_TITLE"],
  "steps": {
    "build": {
      "commands": ["...", "./check-env.sh"]
    }
  },
  "deploy": { "startCommand": "node main.js" }
}
```

Same build command, output:

```
DATABASE_URL=NOT SET
SENTRY_AUTH_TOKEN=NOT SET
VITE_APP_TITLE=MyApp
MY_SECRET=NOT SET
```

[...]

PR #460, opened 2026-02-15
fix: preserve --env secrets when config defines secrets

cadeljones:
## Summary
- Preserve CLI `--env` secrets when config file also defines secrets by merging both sources instead of replacing one with the other.
- Add coverage for secret merge behavior in `core/core_test.go` to prevent regressions.

## Context
- Issue #455


iloveitaly:
@cadeljones thanks for this—great fix. Two things I need to do before I can merge this:

1. Add an integration test for this (examples/ entry that tests this functionality)
2. Update snapshots (`mise run test-update-snapshots`)

If you have time to do those, great! If not, I'll get them when I can.

Issue #419, opened 2026-01-11
List extra packages from mise.toml in output


aleksrutins:
When more packages are defined in `mise.toml` than are required by the provider, they're installed, but not listed at the top of the output.

<img width="449" height="193" alt="Image" src="https://github.com/user-attachments/assets/2b800e6c-002a-44e8-a36a-76dccef731d2" />

<img width="375" height="131" alt="Image" src="https://github.com/user-attachments/assets/f20bcfbc-c329-4540-a8b6-9d6f4ab14d3c" />

iloveitaly:
This is a great callout, thanks!

Including all of the packages may be verbose, but adding a line indicating that other packages specified in mise.toml will be installed feels like a improvement here.

manasa-bhagwat:
Hi @iloveitaly,

I'd like to work on this.

From the discussion it sounds like the goal is not to list every extra tool from `mise.toml`, but instead add a short informational line in the output indicating that additional tools defined in `mise.toml` will also be installed.

I'll try to reproduce the behavior locally and open a PR with this approach if no one else is already working on it.

iloveitaly:
@manasa-bhagwat awesome, would love to have your help! Yes, that's right.

Even better, maybe we could list out 2-3 tools and then `, etc` is there are more than 2-3 installed.

Sound good?

manasa-bhagwat:
> [@manasa-bhagwat](https://github.com/manasa-bhagwat) awesome, would love to have your help! Yes, that's right.
> 
> Even better, maybe we could list out 2-3 tools and then `, etc` is there are more than 2-3 installed.
> 
> Sound good?

Thanks @iloveitaly, that makes sense!

I'll implement it so that if additional tools are detected from `mise.toml`, the output lists up to 2–3 of them and shows `, etc.` if there are more.

I'll try reproducing the behavior locally and open a PR with this approach shortly.

iloveitaly:
Awesome, thanks!

manasa-bhagwat:
@iloveitaly 

I’ve opened a PR for #419 with the output update for additional `mise.toml` tools (up to 2 names, then `, etc.` when more). 

Please review when you have a moment.

PR #496, opened 2026-03-10
fix: show additional mise.toml tools in output summary

manasa-bhagwat:
This PR addresses #419, adds test coverage for the new output behavior, and passes `mise run check` and `mise run test`.

## Summary
- Add an informational output line when additional tools are detected from `mise.toml` beyond provider-resolved packages.
- Show up to 3 additional tool names and append `, etc.` when there are more.
- Keep the existing `Packages` table focused on resolved/provider packages.

## Reproduction (before)
- Ran:
  - `mise run cli -- --verbose build --show-plan examples/mise-config`
- `examples/mise-config/mise.toml` defines `node`, `bun`, `python`, `go`.
- Output only listed `node` and `bun` in `Packages`, with no mention of extra tools.

## After
- Output now includes:
  - `Also installing additional tools from mise.toml: go, python`

list packages

## Test Plan
- `mise run check`
- `mise run test`
- Manual validation:
  - `mise run cli -- --verbose build --show-plan examples/mise-config`
  - `mise run cli -- --verbose info examples/mise-config`

### DIFF
--- a/core/generate/mise_step_builder.go
+++ b/core/generate/mise_step_builder.go
@@ -26,8 +26,9 @@ var (
 
 // represents a app-local mise package
 type MisePackageInfo struct {
-	Version string
-	Source  string
+	Version          string
+	RequestedVersion string
+	Source           string
 }
 
 // MiseListSource represents the source of a mise tool installation
@@ -61,6 +62,7 @@ type MiseStepBuilder struct {
 	Variables             map[string]string
 	app                   *a.App
 	env                   *a.Environment
+	cachedMisePackages    map[string]*MisePackageInfo
 }
 
 func (c *GenerateContext) NewMiseStepBuilder(displayName string) *MiseStepBuilder {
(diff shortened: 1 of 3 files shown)


iloveitaly (changes requested):
Thanks so much for this! Left some comments. I love the idea here and would like to get this merged!

manasa-bhagwat:
@iloveitaly I’ve addressed the review feedback and updated the evidence screenshot.

Could you please take another look when you have a moment?.

core/generate/mise_step_builder.go
```diff
@@ -26,8 +26,9 @@ var (
 
 // represents a app-local mise package
 type MisePackageInfo struct {
-	Version string
-	Source  string
+	Version          string
+	RequestedVersion string
```
iloveitaly, line 30: why did we need this?

iloveitaly (changes requested):
We don't want to display _every_ package in mise.toml. It could get a [little crazy](https://github.com/iloveitaly/python-starter-template/blob/master/.config/mise.dev.toml) and most folks who have a lot in their mise config know it's there, so it doesn't add much value to display it.

I think what we should do here is:

1. Limit the number of mise packages we display to 5
2. This is display-only logic, so we if we can keep it contained to a single method which makes it clear this only for displaying more info to users that would be great.

PR #518, opened 2026-03-16
feat: add --cache-ref flag for registry-based BuildKit layer cache

ccc-uglydawg:
## Summary
Adds a `--cache-ref` CLI flag that configures BuildKit to use a remote container registry for layer cache import/export (`type=registry`).

This enables stateless CI/CD workers to share build cache across deploys by storing layers in a container registry, rather than relying on local disk cache that is lost on pod restart.

## Usage
```bash
railpack build . --cache-ref registry.example.com/cache:latest
```

## Changes
- `cli/build.go`: Add `--cache-ref` string flag, pass to `BuildWithBuildkitClientOptions`
- `buildkit/build.go`: Add `CacheRef` field, configure `CacheImports`/`CacheExports` with `type=registry` and `mode=max`

## Context
BuildKit natively supports registry-based cache via `CacheImports`/`CacheExports` with `type=registry`. Railpack already supports GHA cache (`ImportCache`/`ExportCache` fields with `type=gha`) but doesn't expose registry cache. This is a small addition that unlocks stateless build workers for multi-tenant platforms where each customer has their own registry.

### DIFF
--- a/buildkit/build.go
+++ b/buildkit/build.go
@@ -12,11 +12,13 @@ import (
 
 	"github.com/charmbracelet/log"
 	"github.com/containerd/platforms"
+	"github.com/docker/cli/cli/config"
 	"github.com/moby/buildkit/client"
 	_ "github.com/moby/buildkit/client/connhelper/dockercontainer"
 	_ "github.com/moby/buildkit/client/connhelper/nerdctlcontainer"
 	"github.com/moby/buildkit/client/llb"
 	"github.com/moby/buildkit/session"
+	"github.com/moby/buildkit/session/auth/authprovider"
 	"github.com/moby/buildkit/session/secrets/secretsprovider"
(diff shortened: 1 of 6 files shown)


iloveitaly:
@ccc-uglydawg seems like a great improvement! Do you know if builtkit will throw an error if a bad cache registry is provided or should we do some sort of validation there?

iloveitaly:
@ccc-uglydawg https://github.com/railwayapp/railpack/pull/623 and https://github.com/railwayapp/railpack/pull/595 implemented part of what you have in this PR.

I'm not familiar with the cpln auth stuff. Happy to work on that in a separate PR. If you want to try to upstream that change, could you please provide an integration test or additional information about how to set up a registry locally that uses that authentication type? Some more information about the use case such as which registries use this type of authentication would be helpful too.

Issue #207, opened 2025-08-01
Issue with building node projects


nickpapciak:
Trying to build an app that uses node fails because of bad keys. You have to set `MISE_NODE_VERIFY=false` to get it to work

```bash
 => ERROR install mise packages: node                                                                                  2.7s
------
 > install mise packages: node:
0.121 mise trusted /app
0.409 mise node@22.18.0    install
0.412 mise node@22.18.0    download node-v22.18.0-linux-x64.tar.gz
2.667 gpg: Signature made Thu Jul 31 21:47:23 2025 UTC
2.667 gpg:                using EDDSA key 5BE8A3F6C8A5C01D106C0AD820B1A390B168D356
2.668 gpg: Can't check signature: No public key
2.669 mise ERROR gpg failed
2.669 gpg: Signature made Thu Jul 31 21:47:23 2025 UTC
2.669 gpg:                using EDDSA key 5BE8A3F6C8A5C01D106C0AD820B1A390B168D356
2.669 gpg: Can't check signature: No public key
2.673 mise ERROR failed to install core:node@22.18.0
2.673 mise ERROR gpg exited with non-zero status: exit code 2
2.673 mise ERROR Run with --verbose or MISE_VERBOSE=1 for more information
------
ERROR: failed to solve: process "sh -c mise trust -a && mise install" did not complete successfully: exit code: 1
```

and

```bash
➜ mise install --verbose
DEBUG ARGS: mise install --verbose
DEBUG config: ~/Downloads/df_demo/mise.toml
DEBUG config: ~/.config/mise/config.toml
DEBUG install_some_versions: node@22.18.0
INFO  node@22.18.0    install
INFO  node@22.18.0    download node-v22.18.0-darwin-arm64.tar.gz
DEBUG GET Downloading https://nodejs.org/dist/v22.18.0/node-v22.18.0-darwin-arm64.tar.gz to ~/.local/share/mise/downloads/node/22.18.0/node-v22.18.0-darwin-arm64.tar.gz
DEBUG GET https://nodejs.org/dist/v22.18.0/node-v22.18.0-darwin-arm64.tar.gz
DEBUG starting new connection: https://nodejs.org/
DEBUG GET https://nodejs.org/dist/v22.18.0/node-v22.18.0-darwin-arm64.tar.gz 200 OK
[...]

PR #540, opened 2026-03-28
feat: 14d delay on new mise packages

iloveitaly:
## Motivation

Avoid broken or incomplete tool versions by ensuring only versions released at least 14 days ago are resolved by default. This should help with the python & ruby compilation issues and the node version verification stuff (I'd like to to remove the node verification skip in the future).

## Description

- Set the mise `install_before` setting to `14d` in the default configuration.
- Add a table to the documentation explaining all default mise settings and the reason for each setting.

## Screenshots / Test

Verified by updating the mise configuration snapshot in `core/generate/__snapshots__/context_test.snap`.

## Links

- Relates to #207

### DIFF
--- a/core/generate/__snapshots__/context_test.snap
+++ b/core/generate/__snapshots__/context_test.snap
@@ -28,7 +28,7 @@
  "steps": [
   {
    "assets": {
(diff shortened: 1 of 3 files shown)


iloveitaly:
cc @coffee-cup

PR #546, opened 2026-04-05
feat: surface extra mise.toml packages in build output

### LINKED ISSUE #419: List extra packages from mise.toml in output

constantinexanthos:
Closes #419

## Problem

When users define additional packages in their `mise.toml` (e.g. `redis`, `terraform`), those packages are installed during the build but never shown in the output. The install step only lists packages that Railpack's providers auto-detected, leaving users with no visibility into what else is being installed.

## Solution

After collecting provider-detected packages, the build step now reads the user's mise config files, parses the `[tools]` section, and appends any extra package names (not already tracked by providers) to the install step's display name.

**Before:** `install mise packages: bun, node`  
**After:** `install mise packages: bun, node, go, python`

## Changes

- `core/generate/mise_step_builder.go`: Parse user mise.toml files for extra tool names and include them in build output
- Updated `mise-config` snapshot to reflect the new output

## Testing

- All 110 snapshot tests pass
- Tested with a sample project containing extra mise.toml packages

### DIFF
--- a/core/__snapshots__/TestGenerateBuildPlanForExamples_mise-config_1.snap.json
+++ b/core/__snapshots__/TestGenerateBuildPlanForExamples_mise-config_1.snap.json
@@ -79,7 +79,7 @@
     },
     {
      "cmd": "mise install",
-     "customName": "install mise packages: bun, node"
+     "customName": "install mise packages: bun, node, go, python"
     }
    ],
    "inputs": [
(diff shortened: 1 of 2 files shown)


iloveitaly:
https://github.com/railwayapp/railpack/pull/496 has a better approach: using parsed mise output instead of attempting to parse `mise.toml`. Could you help with further improvements to that PR instead? Going to close this one out.

PR #547, opened 2026-04-05
feat: run containers as non-root user by default

### LINKED ISSUE #286: Runs as root user
Container is running as root user by default

constantinexanthos:
Closes #286

## Problem

Containers built by Railpack run as root by default, which is a security concern and violates container best practices.

## Solution

Containers now run as UID 1001 (`railpack` user) by default. A new `setup:user` build step creates the user/group and sets ownership of `/app`.

**Changes across 5 files:**
- `core/plan/plan.go` — Added `User` field to `Deploy` struct
- `core/config/config.go` — Added `user` option to `DeployConfig` so users can override
- `core/generate/deploy_builder.go` — Defaults to UID `1001`, creates non-root user step with `groupadd`/`useradd`/`chown`
- `core/generate/context.go` — Wires config override to deploy builder
- `buildkit/convert.go` — Sets `User` in OCI image config

**Opt-out:** Users who need root can set their railpack config:
```json
{"deploy": {"user": "root"}}
```

## Testing

- All 110 snapshot tests updated and passing
- Compiles clean on macOS (Apple Silicon)

### DIFF
--- a/core/generate/deploy_builder.go
+++ b/core/generate/deploy_builder.go
@@ -11,6 +11,7 @@ type DeployBuilder struct {
 	Variables    map[string]string
 	Paths        []string
 	AptPackages  []string
+	User         string
 }
 
 func NewDeployBuilder() *DeployBuilder {
@@ -21,6 +22,7 @@ func NewDeployBuilder() *DeployBuilder {
 		Variables:    map[string]string{},
 		Paths:        []string{},
 		AptPackages:  []string{},
+		User:         "1001",
 	}
 }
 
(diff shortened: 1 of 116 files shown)


core/generate/deploy_builder.go
```diff
@@ -66,10 +68,26 @@ func (b *DeployBuilder) Build(p *plan.BuildPlan, options *BuildStepOptions) {
 		baseLayer = plan.NewStepLayer(runtimeAptStep.Name)
 	}
 
+	// Create non-root user unless explicitly set to "root"
+	if b.User != "" && b.User != "root" {
+		userStep := plan.NewStep("setup:user")
+		userStep.Inputs = []plan.Layer{baseLayer}
+		userStep.AddCommands([]plan.Command{
+			plan.NewExecCommand(
+				"groupadd -g 1001 railpack && useradd -u 1001 -g railpack -m -s /bin/bash railpack && chown -R 1001:1001 /app",
```
coffee-cup, line 77: I might be missing something, but this should use `b.User` I think?

PR #558, opened 2026-05-15
docs: Railpack provider values in schema enum

radiantjade:
## Summary
- add provider enum values to the generated `railpack schema` output
- document the available `provider` values in the config file guide
- make configured provider lookup case-insensitive so the documented lowercase values work for `elixir` and `dotnet`

This also clarifies that `uv` support is handled by the `python` provider, not a separate provider.

Related Central Station question: https://station.railway.com/questions/what-are-the-available-providers-in-rail-bae3a690

## Verification
- `git diff --check`
- `go test ./core/config ./core/providers`
- `go run ./cmd/cli schema` shows the provider enum
- `bun run build` in `docs`

I also tried `go test ./...`; it did not complete cleanly in this environment because existing example snapshots resolve newer Caddy/FrankenPHP patch versions and integration tests require `BUILDKIT_HOST`. Those failures appear unrelated to this change.

### DIFF
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -27,7 +27,7 @@ type StepConfig struct {
 }
 
 type Config struct {
-	Provider         *string                `json:"provider,omitempty" jsonschema:"description=The provider to use"`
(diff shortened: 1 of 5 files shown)


iloveitaly:
Great improvement! Thanks.

PR #560, opened 2026-05-15
Preserve exact Gradle wrapper versions

### LINKED ISSUE #554: Java provider: gradle wrapper version is truncated to major-only, causing mise to install milestone previews
## Summary

`core/providers/java/gradle.go`'s `setGradleVersion()` extracts the **full** version string (e.g. `9.5.0`) from `gradle-wrapper.properties`'s `distributionUrl`, but then deliberately discards minor and patch components and passes only the **major** version to mise. Mise resolves "latest 9.x" against its registry and currently returns `9.6.0-milestone-1` (a preview that should never be auto-installed for a build pinning `9.5.0`).

The fix is a one-line change: pass the already-extracted full version string to mise, instead of re-parsing it down to just the major.

## Reproduction

A real Railway build with `gradle-wrapper.properties` pinning `gradle-9.5.0-bin.zip`:

```
gradle-wrapper.properties:
  distributionUrl=https\://services.gradle.org/distributions/gradle-9.5.0-bin.zip
```

Railpack build log Packages summary:

```
Packages
  gradle  │  9.6.0-M1  │  gradle-wrapper.properties (9)
```

The `(9)` source annotation reveals what's happening — Railpack reports the major version as the source signal it sent to mise. Mise then returned `9.6.0-M1` because that's the highest 9.x tag in its (currently buggy) registry index.

The build still produces a working artifact: the gradle wrapper script bootstraps its own pinned `9.5.0` zip on top, and gradle 9.5.0 runs the build (confirmed by `https://docs.gradle.org/9.5.0/userguide/...` in the daemon-startup log line). But the milestone gets installed by mise into the builder image's `/mise/installs/gradle/9.6.0-M1/` (~165 MB extracted) and is never used.

## Root cause — the buggy lines

[`core/providers/java/gradle.go`](https://github.com/railwayapp/railpack/blob/main/core/providers/java/gradle.go) `setGradleVersion()`:

```go
// First regex correctly extracts "9.5.0" from the distributionUrl
versionRegex, err := regexp.Compile(`(distributionUrl[\S].*[gradle])(-)([0-9|\.]*)`)
...
customVersion := string(versionRegex.FindSubmatch([]byte(wrapperProps))[3])
// customVersion is now "9.5.0"

// Second regex deliberately captures ONLY the first numeric group (major)
parseVersionRegex, err := regexp.Compile(`^(?:[\sa-zA-Z-"']*)(\d*)(?:\.*)(\d*)(?:\.*\d*)(?:["']?)$`)
...
parsedVersion := string(parseVersionRegex.FindSubmatch([]byte(customVersion))[1])
// parsedVersion is now "9"  ← throws away minor + patch

miseStep.Version(gradle, parsedVersion, "gradle-wrapper.properties")
// Mise is asked for "9", returns whatever the highest 9.x tag in its index is
```

## Suggested fix

The `customVersion` extracted by the first regex is already exactly the version we want to install. Pass it through:

```go
miseStep.Version(gradle, customVersion, "gradle-wrapper.properties")
```

Drop the second regex entirely. The Packages panel in the build log will then show `gradle │ 9.5.0 │ gradle-wrapper.properties (9.5.0)` — matching what the user actually pinned.

This also matches how the JDK is already resolved elsewhere (the `RAILPACK_JDK_VERSION=25` env var is passed through verbatim, not truncated).

## Why this matters

- **Reproducibility.** Major-version resolution means the installed mise binary changes whenever a new minor / RC / milestone tag appears upstream, even when the project's wrapper is unchanged. That's the opposite of what wrapper-pinning is for.
- **Image size & build time.** The unused milestone install costs ~165 MB in the builder image. (Whether it leaks into the runtime image depends on how the runtime stage's `copy /mise/installs` is scoped — in our build it doesn't, but the builder still pays the cost.)
- **Defense in depth.** There's an upstream mise registry bug ([jdx/mise#9499](https://github.com/jdx/mise/discussions/9499)) where `latest_stable_version` for `gradle` is incorrectly reported as a milestone preview. Even after that's fixed, asking mise for `latest <major>` is structurally less reliable than asking for an exact version. This Railpack fix removes the surface area entirely — it stops mattering whether mise/aqua/vfox correctly filter prereleases for `latest <major>` resolution, because Railpack would no longer use that resolution path.

## Cross-references

- Upstream registry bug at mise's gradle index: [jdx/mise#9499](https://github.com/jdx/mise/discussions/9499)
- Source-of-truth for gradle release dates and prerelease flags: `https://api.github.com/repos/gradle/gradle-distributions/releases` and `https://services.gradle.org/versions/all`

## Out of scope (but nearby)

`setMavenVersion()` may have the same shape if it also reads a wrapper file and truncates to major — worth a quick audit while making the gradle fix.

radiantjade:
Fixes #554.

This passes the full Gradle version pinned in `gradle/wrapper/gradle-wrapper.properties` through to mise instead of truncating it to the major version. That avoids `latest <major>` resolution selecting newer milestones or previews when the wrapper is pinned to an exact stable version.

The wrapper URL parser now captures versions from `gradle-<version>-bin.zip` and `gradle-<version>-all.zip`, including patch and prerelease suffixes.

Verification:
- `go test ./core/providers/java`
- `go test ./core -run 'TestGenerateBuildPlanForExamples/java-gradle'`\n- `git diff --cached --check`

### DIFF
--- a/core/providers/java/gradle.go
+++ b/core/providers/java/gradle.go
@@ -34,29 +34,17 @@ func (p *JavaProvider) setGradleVersion(ctx *generate.GenerateContext) {
 		return
 	}
 
-	versionRegex, err := regexp.Compile(`(distributionUrl[\S].*[gradle])(-)([0-9|\.]*)`)
+	versionRegex, err := regexp.Compile(`gradle-([0-9][0-9A-Za-z.-]*)-(?:bin|all)\.zip`)
 	if err != nil {
 		return
 	}
 
-	if !versionRegex.Match([]byte(wrapperProps)) {
+	matches := versionRegex.FindStringSubmatch(wrapperProps)
+	if len(matches) < 2 {
 		return
 	}
 
-	customVersion := string(versionRegex.FindSubmatch([]byte(wrapperProps))[3])
-
-	parseVersionRegex, err := regexp.Compile(`^(?:[\sa-zA-Z-"']*)(\d*)(?:\.*)(\d*)(?:\.*\d*)(?:["']?)$`)
-	if err != nil {
-		return
-	}
-
-	if !parseVersionRegex.Match([]byte(customVersion)) {
-		return
-	}
-
-	parsedVersion := string(parseVersionRegex.FindSubmatch([]byte(customVersion))[1])
-
-	miseStep.Version(gradle, parsedVersion, "gradle-wrapper.properties")
+	miseStep.Version(gradle, matches[1], "gradle-wrapper.properties")
 }
 
 func (p *JavaProvider) gradleCache(ctx *generate.GenerateContext) string {
--- a/core/providers/java/gradle_test.go
+++ b/core/providers/java/gradle_test.go
@@ -1,0 +1,79 @@
+package java
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/railwayapp/railpack/core/app"
+	"github.com/railwayapp/railpack/core/config"
+	"github.com/railwayapp/railpack/core/generate"
+	"github.com/railwayapp/railpack/core/logger"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSetGradleVersionUsesFullWrapperVersion(t *testing.T) {
+	tests := []struct {
+		name            string
+		distributionURL string
+		expectedVersion string
+	}{
+		{
+			name:            "minor version",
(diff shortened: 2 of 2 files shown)


iloveitaly (changes requested):
thanks for the contribution! Would it be possible to add an example project (`examples/`, we have something for gradle in there already) which covers this logic?

We are a big fan of additional integration tests (via example projects) for this sort of thing.

PR #586, opened 2026-06-06
feat(node): support pnpm binaries in monorepo workspaces

### LINKED ISSUE #584: pnpm doesn't relink monorepo binaries
When you have a monorepo that uses pnpm and has node binaries, it will fail:

Let's say you have `packageA` which has a small `bin` script that runs something like `console.log("Hello")`.

 If you then have `packageB` that depends on `packageA` using `workspace:*` and that package runs `packageA` as it's built script, then it will fail.

I think this is because pnpm installs dependencies before the source files are added (in docker), so it never figures out that `packageA` has a binary that `packageB` will depend on.

I know this is hard to understand, so please let me know if you need a repro (I can cook one up)

iloveitaly:
## Motivation

In pnpm monorepos, builds fail when one workspace package exposes a `bin` and another depends on it via `workspace:*`. Railpack runs `pnpm install` before workspace source files are copied, so pnpm cannot link the binary and downstream build scripts cannot find it.

Fixes #584

## Description

Detect workspace packages with `bin` entries and copy the full workspace layer before install, matching the approach already used for lifecycle scripts and local file dependencies. Secret exposure during install is limited to workspaces that define lifecycle scripts, so binary-only workspaces do not receive secrets unnecessarily.

## Test

- Add `examples/node-pnpm-monorepo-binaries` integration test where `package-b` runs `package-a`'s `hello-cli` bin during build
- Unit tests for `PackageJson.HasBin()` and `Workspace.HasPackageBins()`

### DIFF
--- a/core/providers/node/package_json.go
+++ b/core/providers/node/package_json.go
@@ -19,6 +19,7 @@ type PackageJson struct {
 	DevDependencies map[string]string `json:"devDependencies"`
 	Engines         map[string]string `json:"engines"`
 	Main            string            `json:"main"`
+	Bin             json.RawMessage   `json:"bin"`
 	Workspaces      []string          `json:"-"`
 }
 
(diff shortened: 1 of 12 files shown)


vixalien:
Hello @iloveitaly do you have instructions on how I can test it on my local repo and then maybe we can later get it merged?

PR #595, opened 2026-06-16
feat: respect cache backend type for registry cache import/export, add cache support to `build` subcommand

### LINKED ISSUE #557: Cache images are not pulled from the registry
I have a simple node.js application

```
ls -lah
total 20K
drwxr-sr-x 1 alien alien  168 May  8 17:35 .
drwxr-sr-x 1 alien alien    8 May  8 17:08 ..
-rw-r--r-- 1 alien alien   97 May  8 17:11 app.js
drwxr-sr-x 1 alien alien   82 May  8 17:16 .git
drwxr-sr-x 1 alien alien   58 May  8 17:33 node_modules
-rw-r--r-- 1 alien alien  265 May  8 17:35 package.json
-rw-r--r-- 1 alien alien  114 May  8 17:35 pnpm-lock.yaml
```

```js
// app.js
console.log("Hello, World!");
// sleep forever
setTimeout(() => console.log("never"), 99999999);
```

```json
// package.json
{
  "name": "test",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "start": "node app.js"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "type": "module"
}
```

I have a local docker registry running

```
docker ps
CONTAINER ID  IMAGE                         COMMAND               CREATED     STATUS      PORTS                     NAMES
0fc451ed7ef3  docker.io/library/registry:3  /etc/distribution...  2 days ago  Up 2 hours  127.0.0.1:5555->5000/tcp  registry
```

I prepared the build with

```
railpack prepare .         --plan-out ./railpack-plan.json         --info-out ./railpack-info.json
                   
╭─────────────────╮
│ Railpack 0.23.0 │
╰─────────────────╯

  ↳ Detected Node
  ↳ Using pnpm package manager
            
  Packages  
  ──────────
  node  │  22.22.2  │  railpack default (22)
  pnpm  │  9.15.9   │  railpack default (9)
            
  Steps     
  ──────────
  ▸ install
    $ pnpm add -g node-gyp
    $ pnpm install --frozen-lockfile --prefer-offline
            
  Deploy    
  ──────────
    $ pnpm run start


```

I started a buildkit instance in another window

```bash
rootlesskit buildkit
```

then ran the build

```bash
buildctl   --addr unix:///run/user/$UID/buildkit/buildkitd.sock   --debug   build   --local context=.   --local dockerfile=.   --frontend=gateway.v0   --opt source=ghcr.io/railwayapp/railpack-frontend   --output type=image,name=localhost:5555/vixalien/testing:main,push=false   --import-cache type=registry,ref=localhost:5555/vixalien/testing:main-cache,registry.insecure=true   --export-cache type=registry,ref=localhost:5555/vixalien/testing:main-cache,mode=min,registry.insecure=true
```

The cache is correctly exported to the registry, but is never pulled when the build starts. Is it expected behaviour or should I provide more details?

mvanhorn:
## Summary

Build cache is now imported and exported using the backend type the caller actually requested. A registry cache reference such as `type=registry,ref=localhost:5555/cache` is routed to the `registry` backend instead of being silently forced onto the `gha` backend, so the cache is pulled before the next build instead of every build starting cold.

## Why this matters

In `buildkit/build.go`, both `CacheImports` and `CacheExports` hardcoded `Type: "gha"` regardless of the attrs supplied. A registry cache ref was therefore exported correctly but never imported, which matches the report in #557 ("It seems to push the cache after a build but never pull it before a build"). Separately, the `OutputDir` branch reassigned `solveOpts` wholesale, discarding the `CacheImports`/`CacheExports` configured just above it, so cache import was also lost on the output-dir path.

This change derives the cache backend type from the parsed attrs (honoring an explicit `type=` key, defaulting to `gha` when none is given) for both the import and export entries, and reworks the `OutputDir` branch to mutate the existing `solveOpts` (swapping only the `Exports` entry) so the cache configuration survives. Existing GitHub Actions cache usage is unaffected because `gha` remains the default.

## Testing

- `go build ./buildkit/...`
- `go vet ./buildkit/...`
- `go test ./buildkit/...` (includes new table-driven coverage for registry type extraction, the `gha` default, and empty input)

Fixes #557


iloveitaly:
@mvanhorn thanks for this. The change looks like the right shape. I need to read up on how caching works with buildkit a bit more, so I can't approve this right away, but I was poking at this problem too and came up with a similar shape. 

Do you have any ideas how we could write an integration test for this?

vixalien:
This PR is perfect!!! I think it looks good. Not sure either on how to write the integration test...

mabadir:
Posting this here as well, for visibility: https://github.com/railwayapp/railpack/issues/557#issuecomment-4918059654

mvanhorn:
Thanks @mabadir — good catch on the frontend path. The branch actually covers it now: `frontend.go` parses the `cache-imports` frontend opt and sets `CacheImports` explicitly on its internal solve (see the block around `parseCacheImports` / the solve opts at frontend.go:99-110), since buildkit doesn't auto-apply `--import-cache` to frontend-defined vertices. 18b86b7 also hardens `extractCacheType` to default a missing `type=` to `gha` so the pre-PR default behavior is preserved when no backend type is given.

On CI: the `node-latest-pnpm-mise-native-deps` snapshot failure doesn't reproduce — the test passes locally on this branch with network access; it looks like mise-download flakiness in the runner. And the `integration_tests/http_check.go` `undefined: StringOrArray` build error reproduces identically on a clean base checkout, so it's pre-existing rather than from this PR.

iloveitaly:
> [18b86b7](https://github.com/railwayapp/railpack/commit/18b86b79d4cf61f7463a01d4bd726e2aad41899c) also hardens `extractCacheType` to default a missing `type=` to `gha` so the pre-PR default behavior is preserved when no backend type is given

I'm going to be reverting this change: the cache logic in the build subcommand was completely unused and there's no need to force a default here.

I'll be pushing some final touches and then merging this PR today!

PR #596, opened 2026-06-16
fix: propagate host environment variables to sub-folder contexts

### LINKED ISSUE #591: Bug: RAILPACK_NODE_VERSION is ignored in favor of engines.node when using a sub-folder root
Hi! I may have found a discrepancy between the documentation and the actual behavior of Railpack.

My Railway service is configured with:

- Root Directory: `/backend`
- Environment variable: `RAILPACK_NODE_VERSION=22`

The `/backend/package.json` contains:

```json
{
  "engines": {
    "node": ">=18"
  }
}
```
According to the documentation, RAILPACK_NODE_VERSION should have higher priority than engines.node.

However, during the planning phase Railpack resolves Node.js from package.json:

Packages
──────────
node │ 18.20.8 │ package.json > engines > node (>=18)

To verify that the environment variable exists during build, I added:

echo "RAILPACK=$RAILPACK_NODE_VERSION"

and got:

RAILPACK=22

Expected behavior:

Node 22 should be selected from RAILPACK_NODE_VERSION

Actual behavior:

Node 18.20.8 is selected from package.json > engines.node

I noticed this while using a sub-folder Root Directory (/backend), although I'm not sure whether that is related or merely coincidental.

If needed, I can provide a minimal reproduction repository. 

Image

vunas:
### Overview

Fixes #591.

Sub-folder deployments were ignoring `RAILPACK_NODE_VERSION` because the generated context did not inherit host `RAILPACK_*` variables.

### Root Cause

`GenerateContext` creates an isolated environment for sub-folder evaluation. That environment did not include host configuration variables, causing `GetConfigVariable("NODE_VERSION")` to return an empty value.

As a result, version resolution always fell back to `package.json` engines or provider defaults.

Additionally, Node version resolution allowed `package.json` values to override environment-defined values.

### Changes

- Propagate `RAILPACK_*` variables into sub-folder `GenerateContext` instances.
- Ensure environment-defined Node versions take precedence over `package.json` engine constraints.
- Add defensive environment initialization to avoid nil map access in tests.

### Verification

- Verified `RAILPACK_NODE_VERSION=24` is respected in sub-folder deployments.
- Verified existing test suites pass without regressions.

### DIFF
--- a/core/generate/context.go
+++ b/core/generate/context.go
@@ -3,6 +3,7 @@ package generate
 import (
 	"fmt"
 	"maps"
+	"os"
 	"slices"
 	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


iloveitaly:
Thanks for the PR, but there was some additional refactors that I wanted to make while cleaning this up, so I've done that in this PR: https://github.com/railwayapp/railpack/pull/601

PR #619, opened 2026-07-08
feat(buildkit): attach docker registry auth to solve session

weimzh:
BuildKit keeps no static credential store of its own; without credentials attached to a solve session it cannot authenticate pulls/pushes against private registries. Load the caller's docker CLI config ($DOCKER_CONFIG, default ~/.docker/config.json) and attach an authprovider alongside the existing secrets provider on every SolveOpt.Session.


iloveitaly:
This makes a lot of sense. Thanks for the contribution!

I'm going to fix this up and add some documentation. Should be included in the next release!

iloveitaly:
Closing in favor of https://github.com/railwayapp/railpack/pull/623.

We couldn't force-push updates onto this contributor PR (fork head is `weimzh:main`), so the original work plus maintainer follow-ups (docs, dependency cleanup, and fixing the docker auth session attachables) were rebased onto a new branch for review/merge.

Thanks again @weimzh for the contribution — the original change is preserved and credited in the new PR.

PR #626, opened 2026-07-13
feat(php): add Symfony framework support

### LINKED ISSUE #89: Symfony support
Like it's done for Laravel, it could be possible to detect a Symfony application with the presence of `bin/console` file, which is the equivalent of `artisan` file in a Symfony app.

iloveitaly:
## Motivation

Railpack could build PHP apps but only treated Laravel specially (`public/` root, start-time setup). Symfony apps needed the same zero-config path ([#89](https://github.com/railwayapp/railpack/issues/89)).

## Description

Detect Symfony via `symfony/framework-bundle` (or Flex `bin/console` + `symfony.lock`), set document root to `/app/public`, use prod env defaults, and install framework-specific start scripts (vanilla / Laravel / Symfony) as `/start-container.sh`. Symfony startup runs Doctrine migrations when available (honoring `RAILPACK_SKIP_MIGRATIONS`) and `cache:warmup`; framework choice is recorded as `phpFramework` metadata with a regenerable `examples/php-symfony` integration example.

## Test

- `mise run check`
- `mise run test` (includes PHP plan snapshots)
- `cd examples/php-symfony && mise run test-integration-cwd`

## Links

- Fixes #89


welcoMattic (approved):
LGTM, but I'm not a Go expert.

PR #635, opened 2026-07-19
fix(php): include Mise runtime layer for one package

### LINKED ISSUE #634: PHP provider drops explicitly configured Node package from runtime image when there is only one Mise package
## Problem

The PHP provider only includes the Mise layer in the runtime image when more than one package is installed:

```go
func (p *PhpProvider) ConditionallyIncludeMise(ctx *generate.GenerateContext) {
    if len(ctx.GetMiseStepBuilder().MisePackages) > 1 {
        ctx.Deploy.AddInputs([]plan.Layer{
            ctx.GetMiseStepBuilder().GetLayer(),
        })
    }
}
```

When Node is the only package, it works during the build but is missing from the final image:

```text
sh: 1: exec: node: not found
```

Adding an unrelated second Mise package makes Node available, which suggests
the condition is off by one.

## Proposed fix

Change:

```go
len(ctx.GetMiseStepBuilder().MisePackages) > 1
```

to:

```go
len(ctx.GetMiseStepBuilder().MisePackages) > 0
```

This ensures the Mise layer is included whenever it contains a package.

jordyvandomselaar:
## Motivation

 PHP applications that use exactly one Mise-managed runtime package currently omit that package from the runtime image.

 ## Description

 Include the Mise layer in PHP deploy images whenever at least one Mise-managed package is required, instead of at least 2.

 ## Test

 - Regenerated and reviewed the Laravel 11 and Laravel 12 React build-plan
   snapshots
 - Verified both plans include `packages:mise` in the deploy inputs
 - `core` and `core/providers/php` tests passed

 ## Links

 Fixes [#634](https://github.com/railwayapp/railpack/issues/634)

### DIFF
--- a/core/__snapshots__/TestGenerateBuildPlanForExamples_php-laravel-11-react_1.snap.json
+++ b/core/__snapshots__/TestGenerateBuildPlanForExamples_php-laravel-11-react_1.snap.json
@@ -32,6 +32,16 @@
   "inputs": [
    {
     "include": [
+     "/mise/shims",
+     "/mise/installs",
+     "/usr/local/bin/mise",
+     "/etc/mise/config.toml",
+     "/root/.local/state/mise"
+    ],
+    "step": "packages:mise"
+   },
+   {
+    "include": [
      "/app/node_modules"
     ],
     "step": "prune:node"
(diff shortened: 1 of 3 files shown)


iloveitaly:
Nice catch! Will include this in the release next week.

Thanks for the contribution!

PR #643, opened 2026-07-22
feat(node): gracefully handle out of sync `package-lock.json`

iloveitaly:
## Motivation

A user's lock file can get out of sync for many different reasons. Most of which are triggered by misc npm bugs. This does not show up in dev and only becomes an issue when a user attempts to build an image and receives this message:

> `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.

## Implementation Choices

Not using `npm ci` introduces the following risks:

* Non-deterministic builds. The lockfile check ensures that the _exact_ package versions are installed that are used locally, in CI, etc. Not asserting this check means the prod env can drift from dev and introduce hard-to-debug problems. The biggest issue here from my POV is users won't even know this is happening.
* Slightly slower builds (`npm install` has to resolve package.json, where `npm ci` just looks at the lockfile). Not a big deal IMHO.

Moving to `npm install` _shouldn't_ not impact users. `npm ci` nukes `node_modules`, but this is not cached so it doesn't matter for us.

Here are two ways we could handle this:

### 1. Use `npm install`

Swap `npm ci` with `npm install`. More builds succeed, users are less frustrated, and our code and the user's mental model of what we are doing remain simple.

**My choice:** this decreases build determinism, which I really don't like, but it's the most simple and least failure-prone.

### 2. Run a custom install script

The only benefit this gives the user is a warning message in their builtkit logs. Very few people will see this and we'd have to bundle this custom (more complex) script into railpack and then one of the layers. Ugh.

I don't think there's any benefit here.

```bash
#!/bin/sh

if [ -f package-lock.json ] || [ -f npm-shrinkwrap.json ]; then
    if ! npm ci; then
        echo >&2 "Warning: lockfile is inconsistent; falling back to npm install."
        echo >&2 "Run npm install locally and commit the updated lockfile."

        rm -rf node_modules
        npm install
    fi
else
    npm install
fi
```

### 3. `npm install` by default, opt-in `npm ci`

Switch the default install script, allow a user to opt-in to `npm ci` via an environment variable config.

I don't think this is worth it. They can just customize the install script using `railpack.json` and we include something about this in the docs.

## Links

* https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable
* https://docs.npmjs.com/cli/v10/commands/npm-ci?v=true

### DIFF
--- a/core/__snapshots__/TestGenerateBuildPlanForExamples_node-angular_1.snap.json
+++ b/core/__snapshots__/TestGenerateBuildPlanForExamples_node-angular_1.snap.json
@@ -105,7 +105,7 @@
      "src": "package.json"
     },
     {
-     "cmd": "npm ci"
+     "cmd": "npm install"
     }
    ],
    "inputs": [
--- a/core/__snapshots__/TestGenerateBuildPlanForExamples_node-astro_1.snap.json
+++ b/core/__snapshots__/TestGenerateBuildPlanForExamples_node-astro_1.snap.json
@@ -109,7 +109,7 @@
      "src": "package.json"
     },
     {
-     "cmd": "npm ci"
+     "cmd": "npm install"
     }
    ],
    "inputs": [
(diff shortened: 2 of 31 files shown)


coffee-cup (approved):

iloveitaly:
I ended up adding a ENV for customizing the install script, the `railpack.json` customization ended up getting ugly for such a simple change.

PR #670, opened 2026-07-31
feat(generate): add default exclude patterns for popular providers

iloveitaly:
## Motivation

Projects without a `.dockerignore` or configured excludes can send unnecessary dependency and cache files into the build context, increasing build overhead.

## Description

Provider-specific defaults now supply sensible exclude patterns when users have not defined their own. Explicit `.dockerignore` and configuration excludes take precedence, while Node, Deno, and Python providers define their respective defaults.

## Breaking Changes

Projects without a `.dockerignore` or configured excludes will now omit `.git`, dependency, and local virtual-environment directories from the build context by default. Builds that intentionally rely on checked-in `node_modules`, `venv`, or `.venv` may be affected; users should add a `.dockerignore` or adjust their exclude configuration as needed.

## Test

- Added coverage for default excludes and user-defined override behavior.
- Validated generated build plans across Node, Deno, and Python examples.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,6 +11,13 @@
 
 <!-- describe the change that was made and why you took this approach -->
 
+## Breaking Changes
+
+<!--
+- describe any user-facing behavior changes that could require action from existing users
+- explain who is affected and how they can migrate or preserve the previous behavior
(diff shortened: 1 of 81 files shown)


iloveitaly:
@coffee-cup this change is a little risky, but I think these sort of smart defaults make a lot of sense for users.

Issue #616, opened 2026-06-29
NodeJS high memory usage when using pnpm with devEngines.packageManager


Guillaume-Docquier:
I have a nodejs monorepo for frontend + backend.

I use pnpm to manage my node and pnpm versions through devEngines.runtime and devEngines.packageManager.

However, since I made that change, the memory usage of the backend went from 200MB to 700MB.

I've narrowed it down to specifically using devEngines.packageManager. I've reverted just that to packageManager, and the memory went back down.

My repository is open source, I have collected my debugging steps and findings here:
- https://github.com/Guillaume-Docquier/text-based-browser-game-1/issues/280

iloveitaly:
Interesting. The pnpm versions specified were the same?

Guillaume-Docquier:
Yes! pnpm 11.9.0

See: 
- https://github.com/Guillaume-Docquier/text-based-browser-game-1/pull/288

iloveitaly:
Ok, thanks for the details. This is very odd–I can't think of an obvious reason why this would occur, since I would imagine the memory usage would be driven by the node version and not pnpm and especially not the way that pnpm is installed.

What's interesting is `devEngines` isn't "supported" in [railpack natively via mise](https://github.com/railwayapp/railpack/blob/69427945fe72e40ed92ffaf00a294a07f1ae1933/core/mise/mise.go#L24).  So that _could_ be part of the issue (maybe pnpm and mise are installing conflicting node versions?). In any case, this is a great report and I'll figure out what is going on here!

Guillaume-Docquier:
Yes I found that odd as well. My first investigation steps were around the node version.

Btw, devEngines is supported by corepack! (But railpack doesn't use it)

I'm curious how Railway measures the memory. What Railway reports is higher than `process.memoryUsage().rss` by about 50MB in my current setup. Could it be that the allocated memory is affected by the container build or something?

I'll put back devEngines and compare with `process.memoryUsage()`

Guillaume-Docquier:
I haven't put back devEngines, but right now `process.memoryUsage()` reports 118MB RSS, while Railway reports (and charges) me for 198MB

iloveitaly:
> I haven't put back devEngines, but right now `process.memoryUsage()` reports 118MB RSS, while Railway reports (and charges) me for 198MB

I'd contact railway support about this. My guess is the container itself has some overhead / there are other processes sucking up memory.

iloveitaly:
[...]

PR #678, opened 2026-08-02
feat(mise): support pnpm idiomatic version files

iloveitaly:
## Motivation

Mise can resolve pnpm versions from `package.json`, including `devEngines.packageManager`, but Railpack did not enable pnpm idiomatic version files. Railpack also duplicated part of this parsing and installed pnpm through Corepack when `packageManager` was present.

## Description

- Enable pnpm idiomatic version-file parsing alongside existing tools.
- **Add support for `devEngines.packageManager`**, including the array format (`[{ "name": "pnpm", "version": "..." }]`) and object format, which were previously unsupported.
- Stop installing pnpm through Corepack. Pnpm is now always installed through Mise, including when `package.json` contains `packageManager`.
- Delegate `devEngines.packageManager` and `packageManager` pnpm version parsing to Mise, replacing Railpack's duplicate custom string splitting and `+hash` suffix handling.
- Retain Railpack's package-manager detection along with `engines.pnpm` and lockfile version fallbacks since Mise does not parse those sources.
- Document the exact pnpm version precedence and supported fields.

Corepack remains in use for npm and Yarn projects selected through `packageManager`; this change only moves pnpm installation to Mise.

## Breaking Changes

* **Node.js Version Precedence:** `devEngines.runtime` (when `name` is `"node"`) is now parsed by Mise and takes priority over `engines.node`, `.nvmrc`, `.node-version`, and `mise.toml`. If your `package.json` specifies a `devEngines.runtime` version that differs from `engines.node` or `.nvmrc`, Railpack will now build using the `devEngines.runtime` version.
* **pnpm Version Precedence & Detection:** `devEngines.packageManager` (both object and array formats) is now resolved before `engines.pnpm` or lockfile fallbacks. If defined, it will also select `pnpm` as the package manager before inspecting lockfiles.
* **Corepack Removed for pnpm:** pnpm is now provisioned directly via Mise (`mise install`). Railpack no longer executes Corepack activation commands or sets `COREPACK_HOME` for pnpm builds.

## Test

- Verified array-form `devEngines.packageManager` selects and installs the requested pnpm version in a workspace application.
- Verified object-form `devEngines.packageManager` selects and installs the requested pnpm and node versions.
- Verified a `packageManager` pnpm version with a `+hash` suffix is installed by Mise at the stripped version and the resulting application builds and runs.
- Verified a runtime native-module application works with its Mise-installed pnpm version without Corepack.

## Links

- [Mise idiomatic version files](https://mise.jdx.dev/configuration.html#idiomatic-version-files)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -13,7 +13,11 @@
 
 ## Test
 
-<!-- how is this change tested? Do not indicate that snapshots were updated -->
+<!--
+- Describe only targeted validation that specifically exercises this change.
+- Do not state or imply that all unit tests were run.
+- Do not indicate that snapshots were updated.
+-->
 
 ## Links
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -88,3 +88,9 @@ There are normal unit tests, snapshot tests, and integration tests. The integrat
 - Markdown files in @docs/src/content/docs/ should be limited to 80 columns
 - Do not fix indentation or formatting manually. This is corrected automatically using `mise run check`
 - NEVER commit language-specific cache or build artifacts (e.g. Python `__pycache__/`, `*.pyc`, `*.pyo` files). If you encounter build artifacts that are not excluded by .gitignore, add appropriate patterns to .gitignore to exclude them
+
+# Mise Functionality
+
+If work requires deep knowledge of mise internals, clone the project into `tmp/mise` and investigate the source code.
+
+https://github.com/jdx/mise
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -85,7 +85,6 @@ There are normal unit tests, snapshot tests, and integration tests. The integrat
 
 # File Conventions
 
-- Markdown files in @docs/src/content/docs/ should be limited to 80 columns
 - Do not fix indentation or formatting manually. This is corrected automatically using `mise run check`
 - NEVER commit language-specific cache or build artifacts (e.g. Python `__pycache__/`, `*.pyc`, `*.pyo` files). If you encounter build artifacts that are not excluded by .gitignore, add appropriate patterns to .gitignore to exclude them
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -93,3 +93,7 @@ There are normal unit tests, snapshot tests, and integration tests. The integrat
 If work requires deep knowledge of mise internals, clone the project into `tmp/mise` and investigate the source code.
 
 https://github.com/jdx/mise
+
+## Pull Requests
+
+When writing a pull request add a "---\nPlease remove this to indicate you've reviewed and deslopped your PR description"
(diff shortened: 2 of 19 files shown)


yigiterturk-dev:
Delegating the parsing to Mise is cleaner than what I had in #703 — Mise already reads both `devEngines` entries, so Railpack does not need its own parser:

```
$ cat package.json
{ "devEngines": { "runtime": { "name": "node", "version": "22.11.0" },
                  "packageManager": { "name": "pnpm", "version": "10.4.1" } } }

$ mise ls --current
node  22.11.0  package.json
pnpm  10.4.1   package.json
```

One gap is left on this branch, though: detection still does not read `devEngines`, so a project that declares its package manager only there is planned as npm.

```
$ cat /tmp/de-test/package.json
{
  "name": "de-test",
  "private": true,
  "scripts": { "start": "node index.js" },
  "devEngines": {
    "runtime": { "name": "node", "version": "22.11.0" },
    "packageManager": { "name": "pnpm", "version": "10.4.1" }
  }
}

$ mise run cli plan /tmp/de-test
[install] npm install
```

No lockfile and no `packageManager` field, so `getPackageManager` falls through to the npm default. Mise resolving the pnpm version does not help, because nothing asks for pnpm in the first place.

That matters for #616 specifically: the reporter moved from `packageManager` to `devEngines.packageManager`, which is exactly the case that stops being detected. `getPackageManager` reading `devEngines.packageManager` between the `packageManager` field and the lockfile checks would cover it, and the version would still come from Mise as this PR does.

iloveitaly:
Great catch! Thanks for this. Working on a fix for it.

iloveitaly:
@yigiterturk-dev This is fixed up! Give it a shot and let me know what you think.

Hoping to release this later this week / early next week.

yigiterturk-dev:
Tested 6b6b786 — the devEngines-only case now resolves correctly. My original repro (no lockfile, no `packageManager` field) plans `pnpm install` with node 22.11.0 and pnpm 10.4.1.

Also checked:

- array-form `devEngines.packageManager` → pnpm
- semver ranges (`^10.4.1`) → pnpm
- yarn and bun via `devEngines` → correct manager
- `yarn.lock` present + `devEngines` pnpm → pnpm wins, consistent with how the `packageManager` field already behaves
- `packageManager: npm@10` + `devEngines` pnpm → npm wins, explicit field takes precedence as expected
- `devEngines.runtime` only → still npm, no regression

Full `./core/...` suite is green (20 packages).

One thing I hit while testing: detection silently falls back to npm when the app path goes through a symlink. Same directory, two paths:

```
~/tmp/realdir            → pnpm install
~/tmp/linkdir (symlink)  → npm install
```

The whole mise idiomatic resolution seems to drop out there — node also falls back to the default version instead of 22.11.0 — so it predates this PR and shouldn't block the release, since `/app` is a real path in the build environment. Flagging it only because the fallback is silent, nothing is logged.

Looks good to ship from my side.

PR #695, opened 2026-08-15
fix(build): correct configuration precedence order

iloveitaly:
## Motivation

Configuration sources were merged in reverse priority, allowing `railpack.json` to unexpectedly override environment and CLI settings.

## Description

- Apply configuration precedence in this order: CLI options, environment variables, then `railpack.json`.
- Apply the same precedence when selecting a configuration file.
- Combine and deduplicate secret names across configuration sources.
- Document precedence rules for commands, arrays, packages, Procfiles, and provider defaults.

## Breaking Changes

Projects that define the same option in multiple configuration sources may produce different build plans.

- `--build-cmd` and `--start-cmd` now override corresponding environment and file settings.
- `RAILPACK_INSTALL_CMD`, `RAILPACK_BUILD_CMD`, and `RAILPACK_START_CMD` now override commands in `railpack.json`.
- `RAILPACK_PACKAGES` now determines the version when the same package is declared in `railpack.json`; packages unique to either source remain included.
- `RAILPACK_BUILD_APT_PACKAGES` and `RAILPACK_DEPLOY_APT_PACKAGES` replace corresponding file lists. Include all desired packages—and `...` where applicable—in the higher-priority environment value.
- `--config-file` now overrides `RAILPACK_CONFIG_FILE` when both are provided, which may select a different file or surface a missing-file error.

Projects that configure each option in only one source are unaffected. Empty CLI and environment values remain ignored; remove stale higher-priority settings to fall back to `railpack.json`.

## Test

- Added unit coverage for CLI, environment, file, Procfile, and provider-default precedence.
- Added coverage for configuration-file selection and empty values.
- Validated precedence through the `config-precedence` integration example.

### DIFF
--- a/buildkit/frontend.go
+++ b/buildkit/frontend.go
@@ -70,6 +70,7 @@ func Build(ctx context.Context, c client.Client) (*client.Result, error) {
 		return nil, err
 	}
 
+	// TODO why are we marshalling the plan here if we don't do anything with it?
 	_, err = json.MarshalIndent(plan, "", "  ")
 	if err != nil {
 		return nil, fmt.Errorf("error marshalling plan: %w", err)
--- a/cli/build.go
+++ b/cli/build.go
@@ -86,6 +86,7 @@ var BuildCommand = &cli.Command{
 		}
 
 		if cmd.Bool("show-plan") && !cmd.Bool("dump-llb") {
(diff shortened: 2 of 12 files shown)


iloveitaly:
Closing to recreate branch stack under -v2

PR #699, opened 2026-08-15
feat(secrets): promote step secrets into the build plan

iloveitaly:
## Motivation

Using a secret in a build step currently requires declaring it twice:

1. In the top-level `secrets` array so the value is mounted
2. In the step’s `secrets` array so changes invalidate that step’s cache

A step-level declaration by itself does not mount or validate the secret. This
duplication is easy to miss and can silently leave a step without the value it
expects.

## Description

- Promote explicit secret names from retained steps into the generated plan’s
  top-level `secrets` array
- Resolve secrets after plan normalization so removed steps do not contribute
  unused names
- Preserve and deduplicate existing top-level secret declarations
- Resolve `"*"` to secrets declared in the root `railpack.json` `secrets` array
  without including names declared only on other steps
- Preserve `"*"` in `railpack-plan.json` when no concrete secret names exist so
  the frontend can retain full `secrets-hash` cache invalidation
- Add focused tests for promotion, deduplication, wildcard handling, removed
  steps, and mutation safety
- Add a `secrets` field to integration-test `test.json` files for supplying
  BuildKit-only values
- Update the secrets example to exercise step-level declarations and `"*"`
- Document the existing distinction between secret mounting and layer-cache
  invalidation
- Document frontend secret filtering, `secrets-hash` behavior, and GitHub token
  arguments for Docker and `buildctl`
- Remove unused plan serialization from the BuildKit frontend

## Breaking Changes

- A secret declared only on a step is now promoted into the generated plan’s
  top-level `secrets` array. Its value must therefore be supplied during the
  build. A missing step-only secret previously did not fail because it was
  neither mounted nor validated; it now causes the build to fail.
- The frontend already mounts every top-level plan secret into every exec
  command. Consequently, a newly promoted step secret becomes available to all
  build commands, not only the step that declared it. Step-level `secrets`
  arrays continue to control cache invalidation rather than access.
- During plan generation, `"*"` now resolves only to secrets declared in the
  root `railpack.json` `secrets` array. A step using `"*"` will not be
  invalidated when a secret declared solely on another step changes.
- Generated `railpack-plan.json` files may contain concrete secret names where
  they previously contained `"*"`. Manually authored plans containing `"*"`
  retain their existing full-hash invalidation behavior.

## Test

- `mise run check`
- `mise run test`
- `mise run docs-build`
- Added unit coverage in `core/generate/secrets_test.go`
- Added end-to-end coverage through the `examples/secrets` integration fixture
- Attempted the focused `examples/secrets` integration test against the remote
  BuildKit host. It was blocked before executing the example because the
  required Railpack images had no manifest matching the requested platform.

## Links

- N/A

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -28,6 +28,9 @@ Follow these instructions carefully when writing code:
 - Use appropriate error handling with proper error wrapping
 - Do not write comments that are obvious from the code itself; focus on
   explaining why something is done, not what it does
+- Do not end one-line comments with a period
+- Do not preserve behavior or add complexity solely to avoid snapshot changes;
+  update and review snapshots when intended behavior changes
 - Seriously, do not write comments that are obvious from the code itself.
 - Do not write one-line functions
 - Always use the App abstraction for file system operations.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,9 +17,13 @@ it into a container image. It's built on BuildKit with support for Node, Python,
 
 Follow these instructions carefully when writing code:
 
-- When writing a comment describing a function, do not start the comment with the name of the function
+- Add a concise one-line comment above functions whose purpose is not
+  immediately obvious. Describe their role or why they exist without starting
+  the comment with the function name.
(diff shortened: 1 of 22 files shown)


iloveitaly:
Closing to recreate branch stack under -v2

PR #703, opened 2026-08-17
feat(node): support devEngines for the runtime and package manager

### LINKED ISSUE #616: NodeJS high memory usage when using pnpm with devEngines.packageManager

yigiterturk-dev:
## Motivation

Railpack does not read `package.json > devEngines` at all, so a project that
declares its Node version and package manager there falls through to lockfile
inference and the Railpack defaults.

In #616 that mismatch is visible at runtime: `devEngines.runtime` asks for a
Node version that Railpack does not install, so mise downloads it when the
container starts and the extracted toolchain stays in the running container's
memory. The reporter measured backend memory going from ~200 MB to ~700 MB
after switching to `devEngines.packageManager`, and back down after reverting.

## Description

`devEngines.runtime` and `devEngines.packageManager` are now read, in both the
single-object and array forms that the [Node.js
spec](https://nodejs.org/api/packages.html#devengines) allows.

Precedence follows the existing pattern of preferring explicit declarations:

- `devEngines.runtime` (node) wins over `engines.node`, and `RAILPACK_NODE_VERSION` still wins over both.
- `devEngines.packageManager` is preferred over lockfile inference, the same way the `packageManager` field already is, and the `packageManager` field still wins over it.
- The declared package manager version is passed to mise, labelled `package.json > devEngines > packageManager`, so it appears in the build log next to the other sources.

## Test

- `TestDevEngines` covers the parser: single objects, arrays, a non-node runtime, and surrounding whitespace.
- A provider test asserts that `examples/node-dev-engines` resolves node `22.11.0` and pnpm `10.4.1`. That example declares no lockfile and no `packageManager` field, so before this change it fell back to npm and the Railpack default Node version.
- `mise run check` and `mise run test` pass. No existing example's build plan changed.

## Links

- Fixes #616
- https://nodejs.org/api/packages.html#devengines

### DIFF
--- a/core/__snapshots__/TestGenerateBuildPlanForExamples_node-dev-engines_1.snap.json
+++ b/core/__snapshots__/TestGenerateBuildPlanForExamples_node-dev-engines_1.snap.json
@@ -1,0 +1,178 @@
+{
+ "caches": {
+  "apt": {
+   "directory": "/var/cache/apt",
+   "type": "locked"
+  },
+  "apt-lists": {
+   "directory": "/var/lib/apt/lists",
+   "type": "locked"
+  },
+  "node-modules": {
+   "directory": "/app/node_modules/.cache",
+   "type": "shared"
+  },
+  "pnpm-install": {
+   "directory": "/opt/pnpm/store",
+   "type": "shared"
+  }
+ },
+ "deploy": {
+  "base": {
+   "step": "packages:apt:runtime"
+  },
+  "inputs": [
+   {
+    "include": [
+     "/mise/shims",
+     "/mise/installs",
+     "/usr/local/bin/mise",
+     "/etc/mise/config.toml",
+     "/root/.local/state/mise"
+    ],
+    "step": "packages:mise"
+   },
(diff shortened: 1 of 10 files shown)


iloveitaly:
Thanks for the contribution. Work to support this is actually already in flight. This is going to be supported through the mise idiomatic file parsing logic.

https://github.com/railwayapp/railpack/pull/678

You can see the pnpm implementation here and the npm implementation is going to look very similar. I'm going to close this out, but let me know if you have any comments on that PR in the general approach.

PR #704, opened 2026-08-17
feat(zig): add a Zig provider

### LINKED ISSUE #674: support zig
I don't know much about this ecosystem. If anyone wants to use Railpack with Zig, let me know / submit a PR.

yigiterturk-dev:
## Motivation

#674 asks for Zig support and invites a PR, since nobody on the team works in
that ecosystem.

## Description

A provider that detects `build.zig`, builds with `zig build --release=safe`, and
runs the executable that `zig build` installs into `zig-out/bin`. Only that
directory is copied into the final image, so the source tree and the build cache
are not shipped.

The executable name comes from `.name` in `build.zig.zon`. That field is an enum
literal since Zig 0.14 (`.name = .my_app`) and a string before that
(`.name = "my_app"`), so both spellings are parsed, and commented-out lines are
skipped because the generated `build.zig.zon` documents the field in comments
above it. When the file is absent, the app directory name is used.

`RAILPACK_ZIG_RELEASE_MODE` selects the release mode (`safe`, `fast`, `small`)
and defaults to `safe`.

Zig is a core mise plugin, so no extra tooling is needed.

## Test

- `examples/zig` is a real project that builds and runs. Verified with Zig 0.16 locally: `zig build --release=safe` produces `zig-out/bin/zig_example`, which prints `Hello from Zig 0.16.0`.
- `TestParseZonName` covers the enum-literal form, the string form, the commented-out field, and a file with no name.
- A provider test asserts detection and the resulting start command, and that a Go project is not detected as Zig.
- `mise run check` and `mise run test` pass. No existing example's build plan changed.

## Links

- Closes #674
- https://ziglang.org/documentation/master/#Zig-Build-System


iloveitaly:
@yigiterturk-dev I'm not familiar with zig. Is there a canonical template for web projects? Could you point me to it?

I'd prefer if we add a `generate.sh` to the example project that uses a generator to build a template from a standarized example.

yigiterturk-dev:
There isn't a web-specific template in Zig core. The canonical scaffold is `zig init`, which ships inside the compiler (`lib/init/`) and produces `build.zig`, `build.zig.zon`, `src/main.zig` and `src/root.zig`. Web servers in Zig are all third-party (`http.zig`, `zap`, `jetzig`), so there's nothing upstream I could point you at for that.

For what it's worth, this example follows the convention the repo already uses for compiled languages — `go-mod`, `cpp-cmake`, `gleam` and `java-maven` all assert on a printed line rather than an HTTP check, and `httpCheck` is reserved for examples whose framework is inherently a web server (Phoenix, Next, Nuxt, dotnet-api).

I've added `examples/zig/generate.sh` as you suggested. It runs `zig init` and copies the result in, so the example is derived from the upstream template instead of being hand-maintained; after a Zig upgrade, re-running it surfaces the changes as a diff. Only `src/main.zig` is replaced, with a minimal program that prints the Zig version — the generated one pulls in the sample library module and demo tests, which are noise here.

Two things the script has to handle, in case they look odd in review:

- `zig init` derives the package, module and binary name from the directory it runs in and refuses to run in a non-empty one. It's invoked inside a fixed-name subdirectory, otherwise the `mktemp` name ends up in `build.zig.zon` and the binary path changes on every run.
- The fingerprint is a permanent package identity that Zig mints once and that must never change afterwards, but `zig init` generates a fresh one each run, so the script carries the existing one over.

That also fixes the CI failure. The example used `std.debug.print`, which writes to stderr, while the integration test asserts on stdout — the program built and printed the expected text, just on the wrong stream. It now uses the same `Io.File.Writer` idiom the `zig init` template itself demonstrates, including the `flush()`.

Verified locally against Zig 0.16.0 (the version CI resolves): `generate.sh` is idempotent, `zig build` succeeds, the binary prints `Hello from Zig 0.16.0` to stdout with stderr empty, `go test ./core/...` passes and the build-plan snapshot is unchanged.

yigiterturk-dev:
@iloveitaly gentle bump on this one whenever you get a moment. The `generate.sh` you asked for is in, and the CI failure is fixed too — the example was printing via `std.debug.print`, which goes to stderr, while the integration test asserts on stdout. No rush, just flagging it so it doesn't get lost.

PR #712, opened 2026-08-19
fix(node): skip install files excluded from the build context

### LINKED ISSUE #707: fix: planner should respect .dockerignore
https://station.railway.com/questions/railpack-mise-2026-8-4-fails-on-a-packag-9f15d8f3

yigiterturk-dev:
## Motivation

Fixes #707.

The planner discovers install files by globbing the app directory, while BuildKit only ever loads the filtered context. When `.dockerignore` (or `railpack.json`'s `exclude`) covers a directory containing a `package.json`, the step still emits a `COPY` for it and the build fails late with `failed to compute cache key: ... not found`.

Minimal reproduction:

```
.
├── .dockerignore          # docs/
├── package.json
├── pnpm-lock.yaml
├── index.js
└── docs/marketing/package.json
```

`railpack plan` emits a `COPY` for `docs/marketing/package.json`, which is exactly the path BuildKit never loads. `railpack.json`'s `exclude` produces the same result.

## Description

`PackageManager.SupportingInstallFiles` filtered only `node_modules/`, and the exclude patterns were assembled separately in `GenerateContext.Generate` and handed straight to BuildKit as `buildPlan.Exclude`, so they never constrained the file list the steps were built from.

This adds `GenerateContext.FilterExcludedFiles`, which runs discovered paths through the merged pattern set, and calls it at the end of `SupportingInstallFiles`. `ExcludePatterns` is extracted so `Generate` and the filter share one definition rather than assembling the list twice.

Matching uses `moby/patternmatcher` — already a dependency, and the same matcher BuildKit uses to build the context — via `MatchesOrParentMatches`, so files under an excluded directory are caught and negation patterns keep working. The matcher is compiled lazily because `railpack.json`'s `exclude` is not final when the context is constructed. If compilation fails the files are kept and a warning is logged, so a malformed pattern cannot silently drop a manifest from the install.

The helper is on `GenerateContext` rather than inside the node provider, since any provider that discovers files by globbing has the same exposure.

## Test

`TestFilterExcludedFiles` covers the matcher: simple patterns, files under an excluded directory, negation, `railpack.json` `exclude`, and the no-patterns passthrough.

`TestSupportingInstallFiles_ExcludedFromContext` covers the reported case end to end — an excluded manifest is not copied, a negated one still is, and without a `.dockerignore` nothing changes.

Verified manually with the CLI across all four combinations above. Full `./core/...` suite and `go vet` pass.

## Links

- #707
- https://station.railway.com/questions/railpack-mise-2026-8-4-fails-on-a-packag-9f15d8f3

---

I offered to send this on the issue and started it while waiting — happy to close it if you had already picked this up or would rather take a different approach.


core/generate/context.go
```diff
@@ -32,6 +33,8 @@ type GenerateContext struct {
 	Env             *a.Environment
 	Config          *config.Config
 	dockerignoreCtx *plan.DockerignoreContext
+	excludeMatcher  *patternmatcher.PatternMatcher
```
iloveitaly: this is only used a single time, right? We shouldn't keep it around in the context.

core/providers/node/package_manager.go
```diff
@@ -369,7 +369,10 @@ func (p PackageManager) SupportingInstallFiles(ctx *generate.GenerateContext) []
 		}
 	}
 
-	return allFiles
+	// The globs above read the app directory directly, so they can pick up
+	// manifests that .dockerignore or `exclude` keep out of the build context.
+	// COPYing one of those fails the build, so drop them here.
+	return ctx.FilterExcludedFiles(allFiles)
```
iloveitaly: This doesn't seem right to me. Instead I think the `ctx.App.*` methods should implement the filtering.

iloveitaly (changes requested):
Thank you for the contribution! This is a great start. However, we need to make this work for not just node-related builds but all builds and we need to update some of the example projects or add a new one to cover this case.

yigiterturk-dev:
Thanks for the review — both points were fair.

- Moved the filtering into `App.findMatches`, which backs `FindFiles`, `FindDirectories`, `HasMatch` and `FindFilesWithContent`, so every provider is covered instead of just the node package manager. `package_manager.go` is reverted.
- Dropped `excludeMatcher`, `excludeCompiled` and `FilterExcludedFiles` from `GenerateContext`. The matcher lives on the `App` now and is compiled once in `NewGenerateContext` via `SetExcludePatterns`. The lazy compile went away too — the config is already final at that point.
- Added `examples/node-npm-dockerignore`: an npm app whose `.dockerignore` excludes `docs/`, with an unrelated manifest at `docs/marketing/package.json`. The generated snapshot confirms nothing under `docs/` reaches the plan.

**One behaviour change worth your explicit sign-off.** Since this now covers all glob-based discovery, it also reaches provider detection. Only two providers detect via globs — dotnet (`*.csproj`) and java (`pom.*`, `gradlew`); everything else uses `HasFile`, which I left unfiltered.

Measured it on an app whose `.dockerignore` excludes its own `app.csproj`:

- before: a full .NET plan is produced, and the build fails later at `COPY app.csproj`
- after: no provider is detected and the plan is empty

Both are failures, but the second one fails early instead of deep inside the build. I think that's the right trade, though it is broader than the original bug — happy to scope it to file collection only if you'd rather.

Related inconsistency I did not touch: `HasFile` and `ReadFile` still bypass the filter, so a root manifest excluded by `.dockerignore` is still detected and then fails at COPY. Worth a follow-up?

yigiterturk-dev:
@iloveitaly both review points are addressed and the branch is rebased onto `main`:

- The filtering moved out of the node provider and into `App.findMatches`, which backs `FindFiles`, `FindDirectories`, `HasMatch` and `FindFilesWithContent` — so every provider respects the build context excludes, not just node.
- The single-use field is gone from `generate.Context`.
- `examples/node-npm-dockerignore` covers the case (a `docs/marketing/package.json` excluded by `.dockerignore` that previously leaked into the install step), with a snapshot.

Re-requesting review whenever you have a moment. CI needs a maintainer approval to run on this fork.

examples/node-npm-dockerignore/README.md
```diff
@@ -0,0 +1,12 @@
+# node-npm-dockerignore
```
iloveitaly, line 1: this is a good test! We should change the name though. It's not testing npm, just dockerignore really. `dockerignore-planner` is a better name.

I'd like to also test the `mise.local.toml` case specifically, since that will be a common use case, so we should add a `mise.toml` and `mise.local.toml` here.

core/generate/context.go
```diff
@@ -350,3 +353,12 @@ func (c *GenerateContext) GetAppSource() string {
 func (c *GenerateContext) GetLogger() *logger.Logger {
 	return c.Logger
 }
+
+// ExcludePatterns returns the exclude patterns that will be applied to the
+// build context, merged from .dockerignore and railpack.json's `exclude`.
+func (c *GenerateContext) ExcludePatterns() []string {
```
iloveitaly, line 359: we can use modern go for this now

```
import "slices"

func (c *GenerateContext) ExcludePatterns() []string {
	return slices.Concat(c.dockerignoreCtx.Excludes, c.Config.Exclude)
}
```

iloveitaly (changes requested):
Really close! Couple minor changes and then we'll be ready to merge in.

iloveitaly:
This was an awesome starting point! It needed a lot of additional improvements which I've rolled into https://github.com/railwayapp/railpack/pull/757

PR #728, opened 2026-08-29
feat(resolver): support version constraints on package.json > engines

### LINKED ISSUE #723: fix: support version constraints on package.json > engines
https://station.railway.com/questions/railpack-not-respecting-bun-version-217d903e

This is not something mise supports. We would have to implement version constraints within Railpack if we wanted to support this

yigiterturk-dev:
Closes #723.

## The problem

mise only understands prefix queries (`bun@1`), so `resolveToFuzzyVersion` flattens a constraint down to a prefix before handing it over. That is lossless for the common forms, but not for these:

| `engines` | fuzzy prefix | what mise installs |
|---|---|---|
| `">=1.1.0 <1.3.0"` | `1` | `1.4.0` — upper bound silently ignored |
| `"<2"` | `<2` | mise error, no prefix to query |
| `"20 \|\| 22"` | `20 \|\| 22` | mise error |
| `"~1.2.3"` | `1.2.3` | pinned to the exact patch, `~` means `>=1.2.3 <1.3.0` |
| `"1.1 - 1.2"` | `1.1 - 1.2` | mise error |

The first row is the one from the linked Station thread: the build looks fine and quietly runs a version the app declared it does not support.

## The change

When a requested version carries meaning the prefix form cannot express — an upper bound, a disjunction, a hyphen range, or a tilde range — the resolver lists the available versions and picks the newest one satisfying the constraint, instead of asking mise for a prefix.

Everything the prefix form already expresses faithfully (exact versions, `14.x`, `^20`, `>=22`, `latest`) keeps the cheaper mise-side resolution, so no existing build plan changes and no snapshots move. A version string that is not a constraint we can parse also falls back to the current path rather than erroring.

The enumeration deliberately does not narrow by prefix: a range can straddle majors (`">=1.9 <2.1"`), so any prefix derived from it risks hiding valid matches.

An unsatisfiable constraint now fails with an error that names the version source, rather than installing something out of range:

```
✖ No version of bun satisfies 20 || 22 (from package.json > engines > bun)
```

## Verified with the CLI

Against a scratch project with `engines` set to each form:

| `engines` | before | after |
|---|---|---|
| `bun: ">=1.1.0 <1.3.0"` | `1.4.0` | `1.2.23` |
| `bun: "<2"` | mise error | `1.4.0` |
| `bun: "~1.2.3"` | `1.2.3` | `1.2.23` |
| `bun: "1.1 - 1.2"` | mise error | `1.2.23` |
| `node: "20 \|\| 22"` | mise error | `22.23.2` |
| `bun: "20 \|\| 22"` | mise error | clear "no version satisfies" error |

## Tests

`mise run check` is clean and `core/resolver` passes. Added unit tests covering which strings take the constraint path vs. the existing fuzzy path, the ten constraint forms above, the interaction with `IsVersionAvailable` (the PHP runtime filter), non-semver entries in mise's output, and the empty-result case.

## Unrelated failure

`TestGenerateBuildPlanForExamples/php-vanilla` fails on this branch, but it also fails on a clean `main` — `dunglas/frankenphp` published `php8.4.25-trixie` and the snapshot still pins `php8.4.24-trixie`. I left that snapshot untouched so it does not get mixed into this change.

### DIFF
--- a/core/resolver/resolver.go
+++ b/core/resolver/resolver.go
@@ -77,9 +77,23 @@ func (r *Resolver) ResolvePackages() (map[string]*ResolvedPackage, error) {
 
 		var latestVersion string
 
-		// If there is a custom version validator, we get possible versions and pick the latest one that matches
-		// Ex: this is used with PHP to match against a available runtime available on docker hub
-		if pkg.IsVersionAvailable != nil {
+		// A bounded or disjunctive range (">=1.1.0 <1.3.0", "20 || 22") cannot be expressed as a
+		// mise prefix query, so we enumerate the candidates and apply the constraint ourselves.
+		if constraint, ok := rangeConstraint(pkg.Version); ok {
+			// Enumerate every version rather than narrowing by prefix: a range can straddle
+			// majors (">=1.9 <2.1"), so any prefix we could derive risks hiding valid matches.
+			versions, err := r.mise.GetAllVersions(name, "")
+			if err != nil {
+				return nil, err
+			}
+
+			latestVersion = newestMatching(versions, constraint, pkg.IsVersionAvailable)
+			if latestVersion == "" {
+				return nil, fmt.Errorf("no version of %s satisfies %s (from %s)", name, pkg.Version, pkg.Source)
+			}
+		} else if pkg.IsVersionAvailable != nil {
+			// If there is a custom version validator, we get possible versions and pick the latest one that matches
+			// Ex: this is used with PHP to match against a available runtime available on docker hub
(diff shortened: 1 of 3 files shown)


yigiterturk-dev:
Rebased onto `main`. The earlier red unit-test job was upstream drift, not this change — the `php-vanilla` snapshot still pinned frankenphp `8.4.24` and `main` has since moved to `8.4.25`. Nothing in this branch touches snapshots; it's resolver code and its unit tests only.

